PR #77, opened 2022-03-21
feat(config): generating typescript definitions

ad2ien:


fredericvilcot (dismissed):

nperon (dismissed):

fredericvilcot:
Good catch ! 👀

amimart (changes requested):
Great, I have a remark not on the content but on the form: Actually I don't think your work is related to a feature like your commit suggest it (i.e. `feat(config):`) or the branch name.

Keep in mind that we generate version numbers based on commit messages, a `feat:` will increment the minor version but, considering the actual change only the patch version should be touched.

ad2ien:
> Great, I have a remark not on the content but on the form: Actually I don't think your work is related to a feature like your commit suggest it (i.e. `feat(config):`) or the branch name.
> 
> Keep in mind that we generate version numbers based on commit messages, a `feat:` will increment the minor version but, considering the actual change only the patch version should be touched.

Ah yes better `chore:` then ?

ccamel (approved):

PR #81, opened 2022-03-24
Feat/add logo component

fredericvilcot:
This PR implements `Logo` component in ØKP4 Design System, following specifications : https://github.com/okp4/ui/issues/97


fredericvilcot:
@ccamel waiting for inner issue management before reviewing

src/ui/atoms/logo/Logo.tsx
```diff
@@ -0,0 +1,57 @@
...
+   * The type of the logo, flexible according to use and location.
+   * By default, the standalone logo including the logotype, the logomark and the slogan..
+   */
+  readonly type?: 'logo' | 'logotype' | 'logomark' | 'slogan'
+}>
+
+export const Logo: React.FC<TLogoProps> = ({ size = 'medium', type = 'logo' }: TLogoProps) => {
+  const { theme }: ThemeContextType = useTheme()
+
+  const getElement = useCallback(() => {
```
ccamel, line 31: Why not` useMemo()` since you want to memorize a value rather than a function I think? But from what I understand, I'm not sure it's useful since the calculation performed is not really expensive, isn't it?

ccamel (changes requested):
Good. Some minor remarks though I'd like you to consider if that is ok with you. :relieved:

src/ui/atoms/logo/Logo.tsx
ccamel: @fredericvilcot at your disposal to discuss this point if you wish, there are certainly things I might be missing.

src/ui/atoms/logo/Logo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import './logo.scss'
+
+export type TLogoProps = Readonly<{
+  /**
+   * The size of the logo, must be determined according to its use and location.
+   */
+  readonly size?: 'small' | 'medium' | 'large'
+  /**
+   * The type of the logo, flexible according to use and location.
+   * By default, the standalone logo including the logotype, the logomark and the slogan..
```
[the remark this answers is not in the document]
fredericvilcot: Yes, I agree with that, we must propagate this way of proceeding to all the specifications of the components

src/ui/atoms/logo/Logo.tsx
fredericvilcot: @ccamel As we discussed, the convention here is to avoid a new function to be created when component proceed a new render. 
I propose to use `useCallback` for all the functions / methods created in a React component in order to benefit from memoization mechanisms and to prevent reliance on a global order function.

ccamel: Good. Thx for explanations.

ccamel (approved):
:+1:

ad2ien (approved):

PR #90, opened 2022-03-30
Text input

ad2ien:
https://github.com/okp4/product/issues/468


ad2ien:
I'll strat from scratch without icon management

PR #94, opened 2022-04-01
chore: update storybook

ad2ien:
With a self explaining title


ccamel:
@fredericvilcot Are we ok with that one? (after conflict resolution of course).

Issue #119, opened 2022-04-08
Button component


ccamel:
## Propose a visual feature

![not-ready](https://img.shields.io/badge/status-notready-yellow.svg)
### **Purpose**

The `button` component is at the heart of the design system because it allows the user to express actions within the interface in which he evolves. This is why the design of this component is particularly important to offer the best possible User eXperience.
The `button` component must be sufficiently configurable to fully integrate into the style and philosophy of the ØKP4 design system.

### **Visual**

**Dark theme**

<img width="227" alt="image" src="https://user-images.githubusercontent.com/9574336/162460342-c4c7ad08-f07a-4cd1-8165-ae797248e057.png">

**Light theme**

<img width="222" alt="image" src="https://user-images.githubusercontent.com/9574336/162460472-91c381aa-28ab-484a-9b82-c5c68961ad1c.png">

### **Use cases**

The use cases associated with the `button` component are very numerous and varied, but in most cases allow to mark a **user action**.

### **Specification**
N/A

### **Notes**
N/A

fredericvilcot:
@ccamel 
I will purpose the following specifications, let me know what you think.

**Props**
- `variant?: 'primary' | 'secondary'`: the variant to use.
- `color?: 'primary' | 'secondary' | 'success' | 'warning' | 'error' | 'info'`: the color of the component, regarding the current theme and following semantic color naming.
- `size?: 'small' | 'medium' | 'large'`: the size of the rendered component.
- `disabled?: boolean`: if true, the component is disabled.
- `startIcon?: React.ReactNode`: an icon component placed before the children.
- `endIcon?: React.ReactNode`: an icon component placed after the children.
- `children?: React.ReactNode`: the content of the `button` component.

**State**

N/A

fredericvilcot:
@ErikssonJoakim FYI

fredericvilcot:
@ccamel maybe we could simply a bit by removing `startIcon` and `endIcon` and giving this responsibility to the caller which could provide (or not) an `icon` component directly into `children` element?

ccamel:
@fredericvilcot Yes. I agree. Let's use the possibilities given by the property `children` to let the caller manage the icons.

fredericvilcot:
@ccamel For now, I'll suggest to only implement a `label` to customize button's content and, in a very next future, we will implement icons as `children` elements.

PR #144, opened 2022-04-20
Feat/button enhancement

fredericvilcot:
This PR brings `Button` enhancement, according to following spec : https://github.com/okp4/ui/issues/119


amimart (approved):

nperon (approved):

src/ui/atoms/button/Button.tsx
```diff
@@ -27,24 +37,29 @@ export type TButtonProps = Readonly<{
...
  */
 export const Button: React.FC<TButtonProps> = ({
-  primary = false,
+  variant = 'primary',
   size = 'medium',
+  backgroundColor = 'primary',
+  disabled = false,
   label,
   onClick,
   ...props
```
ccamel, line 46: Could be removed. (`props` has `{}` type as far as I see).

src/ui/atoms/button/Button.tsx
```diff
@@ -27,24 +37,29 @@ export type TButtonProps = Readonly<{
...
+      disabled: !!disabled
     })}
+    disabled={disabled}
     onClick={onClick}
+    title={label}
     type="button"
     {...props}
   >
-    {label}
+    <Typography as="div" color="highlighted-text" fontSize="small" fontWeight="light" noWrap>
```
ccamel, line 61: From what I see, you don't handle child components, so we're not quite compliant with the spec discussed in [ui/issues/119](https://github.com/okp4/ui/issues/119#issuecomment-1100130307) regarding the capabilities to display icons in the button.

ccamel (changes requested):

src/ui/atoms/button/Button.tsx
fredericvilcot: @ccamel 
I updated issue to be compliant with the specification at the moment.

src/ui/atoms/button/Button.tsx
fredericvilcot: Yes you are right, and I realize that this problem is present on the other atomic components and needs to be fixed.
I propose initially to delete this element and thus to partition the use of the button to the properties defined by the interface.
Subsequently, we can discuss whether we extend the type to that of the `HTMLButtonElement`

ccamel: Good to me!

ccamel (approved):

nperon (approved):

amimart (approved):

Issue #120, opened 2022-04-08
Card component


ccamel:
## Propose a visual feature

![not-ready](https://img.shields.io/badge/status-notready-yellow.svg)
### **Purpose**

_TBD_

### **Visual**

**Dark theme**

<img width="327" alt="image" src="https://user-images.githubusercontent.com/9574336/162462879-31707e05-5cbf-4953-9617-4df32912c4c3.png">

**Light theme**

<img width="326" alt="image" src="https://user-images.githubusercontent.com/9574336/162462940-1731936a-da9d-4456-a7c4-239ee01c2a3a.png">

### **Use cases**

_TBD_

### **Specification**

_TBD_

### **Notes**

_TBD_

nperon:
props:

body?: string,
topRightLabel?: string,
bottomLeftLabel?: string,
bottomRightLabel?: string,
size: 'small' | 'medium' | 'large',
background?: 'image1' | 'image2' | 'image3',
onClick?: ()=> void

fredericvilcot:
After review with @nperon , specifications could be : 

- `content?: JSX.Element`,
- `footer?: JSX.Element`,
- `header?: JSX.Element`,
- `size: 'small' | 'medium' | 'large'`,
- `background?:` 'primary' | 'secondary' | 'header' | 'footer'`,
- `withBorder?: boolean`

@ccamel @ErikssonJoakim what do you think ?

ccamel:
Good to me too 👍

PR #146, opened 2022-04-21
Feat/okp4/UI#120 card component

nperon:
This pull request implements specification from issue https://github.com/okp4/ui/issues/120


amimart (changes requested):
Could you please rename your branch to match a `<type>/<name>` format, with `name` containing only lowercase alphanumeric characters or '-' ? For instance `feat/add-card-component` should fit well.

You'll need to close this PR and open a new one.

PR #147, opened 2022-04-21
Feat/error domain core

fredericvilcot:
This PR provides a brand new `error` domain that implements following specifications : https://github.com/okp4/ui/issues/148

### DIFF
--- a/src/domain/error/entity/error.ts
+++ b/src/domain/error/entity/error.ts
@@ -6,7 +6,7 @@ export type Error = Entity<
   {
     readonly name: string
     readonly message: string
-    readonly timeStamp: Date
+    readonly timestamp: Date
     readonly type: string
   },
   Id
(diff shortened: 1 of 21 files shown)


src/domain/error/entity/error.ts
```diff
@@ -0,0 +1,15 @@
+import type { Map } from 'immutable'
+import type { Entity } from 'domain/type/entity.type'
+
+export type Id = string
+export type Error = Entity<
+  {
+    readonly name: string
+    readonly message: string
+    readonly timeStamp: Date
```
amimart, line 9:
```suggestion
    readonly timestamp: Date
```

I think the camel case isn't justified as it is a single word

amimart (changes requested):
Nice tests!

src/domain/error/entity/error.ts
fredericvilcot: Yep, I think it's my eyes ... 👀

ccamel (approved):

amimart (approved):

PR #169, opened 2022-04-28
chore: make engine node compatible with version instead of equivalent to

fredericvilcot:
This PR is pretty straightforward 😃

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "web": "https://okp4.com"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


amimart (approved):
Maybe we can adopt the same strategy concerning the `yarn` version ?

fredericvilcot:
> Maybe we can adopt the same strategy concerning the `yarn` version ?

@amimart Yes, I was planning to do this in a dedicated pull request

Issue #165, opened 2022-04-27
🇺🇸🇫🇷 i18n internationalization


fredericvilcot:
### **Purpose**

All ØKP4 apps must be internationalized, so we need to handle multiple languages as soon as possible.

### **Use cases**

Allows to translate all texts in ØKP4 apps.

### **Specification**

The best way to implement `i18n` is to have a single instance running (singleton) with base translations (`translation_XX.json`) and let each organism declaring its own translation files in `json` format.
For now, available translations will be FR and EN.

### **Acceptance Criteria**

All texts in organisms must be correctly translated (at browser's language switch for example)

### **Notes**

N/A

PR #181, opened 2022-05-02
Feat/i18n

fredericvilcot:
This PR implements following specification : https://github.com/okp4/ui/issues/165


src/i18n/translation_en.json
```diff
@@ -0,0 +1,11 @@
+{
+  "domain": {
+    "error": {
+      "contact-us": "Please try again later or contact us.",
+      "connection-error": "Oops, an error occurred while trying to connect. $t(domain.error.contact-us)",
+      "unspecified-error": "Oops, an unspecified error occurred. $t(domain.error.contact-us)",
+      "validation-error":  "Oops, an error occurred during validation. $t(domain.error.contact-us)",
+      "gateway-error": "Oops, there is a network error. $t(domain.error.contact-us)"
```
ccamel, line 8: You can have gateways that don't use the network. I'd suggest to either introducing a general `network-error` key or changing the message.

src/i18n/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        }
+        return value
+      }
+    },
+    react: {
+      useSuspense: true
+    }
+  })
+
+i18n.addResourceBundle('en', 'translation', translation_en, true)
```
ccamel, line 44: In my opinion, should not be the responsibility the i18n service to register translatations (according to my understanding of the sentence: "let each organism declaring its own translation files in json format." of the spec #181).

src/i18n/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import translation_en from './translation_en.json'
+import translation_fr from './translation_fr.json'
+
+i18n
+  .use(LanguageDetector)
+  .use(initReactI18next)
+  .init({
+    debug: true,
+    detection: {
+      order: ['querystring', 'localStorage', 'navigator', 'htmlTag'],
```
ccamel, line 14: Ideally, language detection behavior should be documented in a storybook. It would be great to have it.

fredericvilcot: I think we could implement it later, but I'm OK with you it would be great to have it so I will take a look for this PR.

src/i18n/index.ts
fredericvilcot: @ccamel 
I'll make the changes 👌

src/i18n/translation_en.json
fredericvilcot: Yes, I will take a look on it

src/i18n/index.ts
fredericvilcot: @ccamel 
I removed resources management from i18n instance initialization and created some utils methods that can register both default resources and any other resource.
These methods should be described in a Storybook as they will be used both by organisms (for registering its own translations files) and by client app (by deciding, or not, to register default translations and any other resource).
Tell me what do you think ?

src/i18n/index.ts
fredericvilcot: Storybook implemented 👌

fredericvilcot:
@ccamel 
I implemented all of your suggestions, please tell me if it's ok for you before merging 🙏

amimart (approved):

ErikssonJoakim (approved):

Issue #212, opened 2022-05-19
Story rendering in Firefox


fredericvilcot:
### **Describe the bug**

In Firefox browser, `stories` are rendered with a tiny height (70px) that make them difficult to read.
This issue is linked to https://github.com/okp4/ui/issues/201: after enabling `features.modernInlineRender`, this issue appeared only in Firefox.

### **To Reproduce**

1. Go to https://okp4.github.io/ui/?path=/docs/atoms-typography--overview
2. Click on **Main Font** story
3. The story is compacted in a 70px height

### **Expected behavior**

Story's height must be dynamic according to its elements size.

### **Screenshots**

![image](https://user-images.githubusercontent.com/47554752/169260529-f8617352-83a0-4172-97d3-7bea906a8e5b.png)

### **Desktop (please complete the following information):**

- OS: MacOs
- Browser Firefox
- Version 100.0.1 (64 bits)

### **Smartphone (please complete the following information):**

N/A

### **Additional context**

N/A

PR #213, opened 2022-05-19
fix(storybook): force height to auto mode for a targetted div

fredericvilcot:
This PR fixes https://github.com/okp4/ui/issues/212


stories/main.scss
```diff
@@ -0,0 +1,7 @@
+.docs-story {
```
ccamel, line 1: Well, if you don't mind, it would be nice to put here a comment to explain the reason (the **why**) for this css trick (possibly with a link to the storybook issue if it exists).

lolottetheclash (approved):
💡

lolottetheclash (approved):

PR #233, opened 2022-05-23
docs: adopt the slashed "O" and ".network" domain

ccamel:
Self explanatory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # OKP4 UI
 
-> Implementation of [OKP4]([okp4.com](https://okp4.com)) UI components in CSS, [Typescript](https://www.typescriptlang.org), and HTML.
+> Implementation of [ØKP4] UI components in CSS, [Typescript](https://www.typescriptlang.org), and HTML.
 
 [![Version](https://img.shields.io/github/v/release/okp4/ui)](https://github.com/okp4/ui/releases)
 [![Lint](https://github.com/okp4/ui/actions/workflows/lint.yml/badge.svg)](https://github.com/okp4/ui/actions/workflows/lint.yml)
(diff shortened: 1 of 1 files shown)


lolottetheclash (dismissed):

fredericvilcot (commented):
Great, maybe we could also modify this [line](https://github.com/okp4/ui/blob/1506772709057a083580e4badb91d9de457e3bd2/LICENSE#L3) in the `LICENSE` file?

amimart (commented):
I think we may also need to update the link in the `package.json`

ccamel:
> Great, maybe we could also modify this [line](https://github.com/okp4/ui/blob/1506772709057a083580e4badb91d9de457e3bd2/LICENSE#L3) in the `LICENSE` file?

I don't think so if we are referring to the company (not the protocol).

fredericvilcot:
> > Great, maybe we could also modify this [line](https://github.com/okp4/ui/blob/1506772709057a083580e4badb91d9de457e3bd2/LICENSE#L3) in the `LICENSE` file?
> 
> I don't think so if we are referring to the company (not the protocol).

I was also referring to the site OKP4.com

ccamel:
> > I don't think so if we are referring to the company (not the protocol).
> 
> I was also referring to the site OKP4.com

Oh, that's right. We need to change it too, I guess.

ErikssonJoakim (approved):

amimart (approved):
Okay seems good !

ErikssonJoakim (approved):

ccamel:
@fredericvilcot Tell me! ☺️

fredericvilcot (approved):
Great 😎

PR #236, opened 2022-05-24
fix(card): increase padding size due to polygon

lolottetheclash:
This PR implements these following specifications: https://github.com/okp4/ui/issues/237
The top left card corner measures 30px soit we have to set the padding by at least this size.

### DIFF
--- a/src/ui/atoms/card/card.scss
+++ b/src/ui/atoms/card/card.scss
@@ -12,9 +12,9 @@ $sizes-map: (
   small-height: 235px,
   medium-height: 235px,
   large-height: 550px,
-  small-padding: 16px,
-  medium-padding: 26px,
-  large-padding: 34px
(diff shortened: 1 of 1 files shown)


fredericvilcot (approved):

ccamel (dismissed):

ErikssonJoakim (changes requested):
For me the padding must be much bigger to get the `header` element out of the polygon. How about if we give the `header` div a padding-left instead? this way we keep the padding for the other elements

ErikssonJoakim (dismissed):
I see now how another PR will solve this issue 😃 👍  Good for me

fredericvilcot (approved):

ErikssonJoakim (approved):

ccamel (approved):

Issue #250, opened 2022-05-30
Card exits viewport


ErikssonJoakim:
### **Describe the bug**

`Card` exits viewport when screen size is smaller than min-width

### **To Reproduce**

Reduce screen size while inspecting [Card](https://okp4.github.io/ui/?path=/docs/atoms-card--overview)

### **Expected behavior**

![Capture d’écran 2022-05-30 à 16 13 11](https://user-images.githubusercontent.com/75730728/171010222-baa23efd-9c9e-4b27-936d-a811c0e838ca.png)

`Card` is expected to adapt to screen size

### **Screenshots**

![Capture d’écran 2022-05-30 à 12 12 32](https://user-images.githubusercontent.com/75730728/171009667-8ee291c2-5faa-4a19-a81d-1e3f5b70a4de.png)

### **Desktop (please complete the following information):**

N/A

### **Smartphone (please complete the following information):**

N/A

### **Additional context**

N/A

PR #252, opened 2022-05-30
Fix/add responsiveness to card

ErikssonJoakim:
This PR implements #250

### DIFF
--- a/src/ui/atoms/card/card.scss
+++ b/src/ui/atoms/card/card.scss
@@ -19,37 +19,36 @@ $sizes-map: (
 
 .okp4-card-main {
   @include with-theme() {
-    clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
-    padding: 1px;
-
-    &.border {
-      background: themed(border);
-    }
-
-    .okp4-card-container {
-      @each $color-key, $color-value in $colors {
-        display: flex;
-        flex-direction: column;
-        justify-content: space-between;
-        gap: 10px;
-        box-sizing: border-box;
+    @each $size in $sizes {
+      &.#{$size} {
         clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
-        max-height: map-get($sizes-map, large-height) - 2px;
-        max-width: map-get($sizes-map, large-width) - 2px;
-        overflow: auto;
-        word-break: break-word;
+        padding: 1px;
+        width: 100%;
+        max-width: map-get($sizes-map, #{$size}-width) - 2px;
 
-        @each $size in $sizes {
-          &.#{$size} {
-            min-height: map-get($sizes-map, #{$size}-height) - 2px;
-            min-width: map-get($sizes-map, #{$size}-width) - 2px;
-            padding-left: map-get($sizes-map, #{$size}-padding);
-            padding-right: map-get($sizes-map, #{$size}-padding);
-          }
+        &.border {
+          background: themed(border);
         }
 
-        &.#{$color-key} {
-          background: themed(#{$color-value});
+        .okp4-card-container {
+          display: flex;
+          flex-direction: column;
+          justify-content: space-between;
+          gap: 10px;
+          box-sizing: border-box;
+          clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
+          max-height: map-get($sizes-map, large-height) - 2px;
+          overflow: auto;
+          word-break: break-word;
+          min-height: map-get($sizes-map, #{$size}-height) - 2px;
+          padding-left: map-get($sizes-map, #{$size}-padding);
+          padding-right: map-get($sizes-map, #{$size}-padding);
+
+          @each $color-key, $color-value in $colors {
+            &.#{$color-key} {
+              background: themed(#{$color-value});
+            }
+          }
         }
       }
     }
--- a/stories/atoms/components/card/Card.stories.mdx
+++ b/stories/atoms/components/card/Card.stories.mdx
@@ -126,8 +126,7 @@ Any component picked from OPK4 UI library e.g. Typography gets styled with color
 
 ## Sizes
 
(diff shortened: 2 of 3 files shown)


src/ui/atoms/card/card.scss
```diff
@@ -19,37 +19,36 @@ $sizes-map: (
...
+    @each $size in $sizes {
+      &.#{$size} {
         clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
-        max-height: map-get($sizes-map, large-height) - 2px;
-        max-width: map-get($sizes-map, large-width) - 2px;
-        overflow: auto;
-        word-break: break-word;
+        padding: 1px;
+        width: 100%;
+        max-width: map-get($sizes-map, #{$size}-width) - 2px;
```
lolottetheclash, line 27: I am not sure that we need this `- 2px` as we discussed yesterday. 
What do you think @fredericvilcot, is there a specific reason?

src/ui/atoms/card/card.scss
```diff
@@ -19,37 +19,36 @@ $sizes-map: (
...
-        gap: 10px;
-        box-sizing: border-box;
+    @each $size in $sizes {
+      &.#{$size} {
         clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
-        max-height: map-get($sizes-map, large-height) - 2px;
-        max-width: map-get($sizes-map, large-width) - 2px;
-        overflow: auto;
-        word-break: break-word;
+        padding: 1px;
```
lolottetheclash, line 25: I think you could move `clip-path`, `padding` and `width` just above `@include with-theme()` because it is not depending on size.

lolottetheclash (changes requested):
Just a few comments but it's a good job! 
Well done!👍

src/ui/atoms/card/card.scss
ErikssonJoakim: You're right I think we can remove all `- 2px`, I don't see how it will change the component other than giving it the `max-width` set by the map

src/ui/atoms/card/card.scss
ErikssonJoakim: Good point I will refactor the code 😃 👍

stories/atoms/components/card/Card.stories.mdx
```diff
@@ -126,8 +126,7 @@ Any component picked from OPK4 UI library e.g. Typography gets styled with color
 
 ## Sizes
 
-Size options are `small`, `medium` and `large`.
-Default is `medium`.
+Size options are `small`, `medium`(default) and `large`. The size also adjust itself according to the screen width
```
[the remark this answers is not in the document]
ErikssonJoakim: Good catch 😄

amimart (dismissed):
I'll place my confidence in our CSS scientists for this one 😉

src/ui/atoms/card/card.scss
```diff
@@ -18,39 +18,39 @@ $sizes-map: (
...
+        .okp4-card-container {
+          display: flex;
+          flex-direction: column;
+          justify-content: space-between;
+          gap: 10px;
+          box-sizing: border-box;
+          clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
+          max-height: map-get($sizes-map, large-height);
+          overflow: auto;
+          word-break: break-word;
```
fredericvilcot, line 43: `word-break` should not be carried by the parent element but rather by the child.

src/ui/atoms/card/card.scss
```diff
@@ -18,39 +18,39 @@ $sizes-map: (
...
+          min-height: map-get($sizes-map, #{$size}-height);
+          padding-left: map-get($sizes-map, #{$size}-padding);
+          padding-right: map-get($sizes-map, #{$size}-padding);
```
fredericvilcot, line 46: We could here simplify using a single line `padding`

fredericvilcot (changes requested):
Responsiveness has been hugely improved, great!
I wrote down some remarks, thanks

src/ui/atoms/card/card.scss
fredericvilcot: 💯 agree

src/ui/atoms/card/card.scss
```diff
@@ -18,39 +18,39 @@ $sizes-map: (
...
-        overflow: auto;
-        word-break: break-word;
+        .okp4-card-container {
+          display: flex;
+          flex-direction: column;
+          justify-content: space-between;
+          gap: 10px;
+          box-sizing: border-box;
+          clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
+          max-height: map-get($sizes-map, large-height);
```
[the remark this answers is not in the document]
ErikssonJoakim: I think the idea was to allow the card to grow if necessary. But I agree cards should keep their proper size

src/ui/atoms/card/card.scss
ErikssonJoakim: 100% right

ccamel (approved):

fredericvilcot (changes requested):

src/ui/atoms/card/card.scss
```diff
@@ -18,38 +18,35 @@ $sizes-map: (
...
+        .okp4-card-container {
+          display: flex;
+          flex-direction: column;
+          justify-content: space-between;
+          gap: 10px;
+          box-sizing: border-box;
+          clip-path: polygon(30px 0, 100% 0, 100% 100%, 0 100%, 0 30px);
+          overflow: auto;
+          max-height: map-get($sizes-map, #{$size}-height);
+          min-height: map-get($sizes-map, #{$size}-height);
```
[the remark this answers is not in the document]
ErikssonJoakim: ✅

fredericvilcot (approved):

lolottetheclash (approved):
It seems good to me ! 👍

Issue #261, opened 2022-06-03
OKP4 UI SSR incompatibility 


lolottetheclash:
## Bug report
Our library is not SSR compatible and limits the use of our components when they use `window` or `document` objects.
### **Describe the bug**
When we import a component from our library, two errors can appear: 
`ReferenceError: document is not defined` or `ReferenceError: window is not defined`.

Our library must be compatible with the SSR mode, and for that we can implement this [solution](https://github.com/vercel/next.js/discussions/35914).

### **Expected behavior**

We should be able to import any component without any errors in SSR mode.

### **Screenshots**

N/A

### **Desktop (please complete the following information):**
N/A

### **Smartphone (please complete the following information):**
N/A

PR #262, opened 2022-06-03
fix: add guard dom access for ssr

fredericvilcot:
This PR implements https://github.com/okp4/ui/issues/261

### DIFF
--- a/src/adapters/wallet/secondary/KeplrWalletGateway.ts
+++ b/src/adapters/wallet/secondary/KeplrWalletGateway.ts
@@ -86,7 +86,8 @@ export class KeplrWalletGateway implements Wallet {
       ) ?? {}
   }
 
-  public readonly isAvailable = (): boolean => !!window.keplr && !!window.keplr.getOfflineSigner
+  public readonly isAvailable = (): boolean =>
+    typeof window !== 'undefined' && !!window.keplr && !!window.keplr.getOfflineSigner
 
   public readonly isConnected = (): boolean => this._isConnected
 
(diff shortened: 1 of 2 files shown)


src/hook/useMediaType.ts
```diff
@@ -9,13 +9,16 @@ export const useMediaType = (query: string): boolean => {
   }, [])
 
   useEffect(() => {
-    const matchQueryList = window.matchMedia(query)
-    setMatches(matchQueryList.matches)
+    // Guard DOM Access for SSR
+    if (typeof window !== 'undefined') {
```
ccamel, line 13: I'm wondering if we couldn't introduce a hook, I mean something like `useIsSSR()` for dealing with the SSR state. What do you think?

src/adapters/wallet/secondary/KeplrWalletGateway.ts
```diff
@@ -86,7 +86,8 @@ export class KeplrWalletGateway implements Wallet {
       ) ?? {}
   }
 
-  public readonly isAvailable = (): boolean => !!window.keplr && !!window.keplr.getOfflineSigner
+  public readonly isAvailable = (): boolean =>
+    typeof window !== 'undefined' && !!window.keplr && !!window.keplr.getOfflineSigner
```
ccamel, line 90: Why not create an explicit predicate named `isBrowser()` or `isSSR` (something meaningful) that tells if `window` is available or not? (since this test being done in several places).

fredericvilcot:
@ccamel 
I close this PR because `SSR` compatibility needs much more work on our `lib` and third-party libraries, I think we should go about it in a different way to solve this problem

PR #281, opened 2022-06-16
ci: fix PR comment publish

ad2ien:
Fix publish action by commenting a pull request:
A git flow or an issue trigger don't end up with the same context. In case of a comment there's a query to do to get the branch of the pull request.

The project is forked so that it's possible to dev on main branch and test the action.

Test with fix/ci-publish branch pull request:
https://github.com/ad2ien/ui/actions/runs/2504934920
Tries to push `ghcr.io/ad2ien/ui:fix-ci-publish` docker image and tries for npm package:
```
🚀 Publishing npm package with following command line: yarn publish --no-git-tag-version --non-interactive --prerelease --preid fix-ci-publish.20220615202551 --tag fix-ci-publish
```

PR #282, opened 2022-06-16
Ci/refine tunings

amimart:
Bring multiple refinements, improvements to CI flows.

## PR from forks

Following https://github.com/okp4/ui/pull/281 some checks were failing due to the secrets available when running actions, when running actions on PR referencing forked head ref we shall avoid referencing organization specific tokens.

**Fix `report-new-dependencies` job:**

Replace the usage of an OKP4 specific token in favor of the default one, which should suffice to comment a PR.

**Fix `lint-dockerfile` job:**

The [luke142367/Docker-Lint-Action](https://github.com/luke142367/Docker-Lint-Action) posting annotations on changes, it needs a token with repository access, which is impossible running from forks. As this linter seems to me less rigid than hadolint, I propose to simply remove it.

## Publish

Publish `next` versions to npm public registry as well, take the opportunity to use matrix job to share the logic with the Github Packages publication.

## Misc

- Prevent the release workflow from running with failing tests.
- Add some concurrency handling.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,10 @@ on:
   pull_request:
     branches: [main]
 
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build-yarn:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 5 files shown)


ad2ien (approved):
Nice! Let's see what it does on a fork...

fredericvilcot (approved):
Thx @amimart for those great improvements! 🚀

ccamel (approved):
👏

PR #288, opened 2022-06-16
Feat/store management

fredericvilcot:
This PR provides the way to build a `redux` store through the OKP4 Design System.

- Each declared store has its own builder which can be invoked anywhere out of the library.
- All builders are tested under unit tests
- All other test suites have been refactored to fully integrate those implementations

👉 The next step will be to export all necessary files to be bundled by `rollup` and then exported by the library.

### DIFF
--- a/src/domain/wallet/usecases/enable-wallet/enableWallet.spec.ts
+++ b/src/domain/wallet/usecases/enable-wallet/enableWallet.spec.ts
@@ -1,7 +1,6 @@
 import { Map, List } from 'immutable'
 import { EventBus } from 'ts-bus'
 import type { ReduxStore } from '../../store/store'
-import { configureStore } from '../../store/store'
 import type { AppState } from '../../store/appState'
 import { enableWallet } from './enableWallet'
 import { InMemoryWalletGateway } from 'adapters/wallet/secondary/InMemoryWalletGateway'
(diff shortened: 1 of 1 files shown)


src/domain/faucet/store/builder/store.builder.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!this.isDependencies(dependencies)) {
+      throw new UnspecifiedError(
+        'Ooops... Dependencies must be provided to build a Faucet store...'
+      )
+    }
+    return new FaucetStoreBuilder({ ...this.faucetStoreParameters, dependencies })
+  }
+
+  public withEventBus(eventBus: DeepReadonly<EventBus>): FaucetStoreBuilder {
+    if (!(eventBus instanceof EventBus)) {
```
lolottetheclash, line 38: As we are using this instance check twice, maybe with could make a function of it? As `isDependencies `?

lolottetheclash (approved):
Nice work, well done !!!

src/domain/faucet/store/builder/store.builder.ts
fredericvilcot: Here we use a simple instanceof based assertion so I'm not sure transforming it to a type-guard is relevant.
isDependencies type guard is much more relevant as it's harder to test 'faucetGateway' in value predicate directly in the if statement.

fredericvilcot: Yes you're right, but I think we could keep this assertion as existing, to mainly prevent from adding a method per assertion in each store builder, if you are OK with that

ccamel (approved):

JoeyDjr (approved):

PR #289, opened 2022-06-16
Feat/add images through base64 urls

lolottetheclash:


fredericvilcot:
@lolottetheclash 
I delete this PR and the branch as all the elements explored in this PR have come to nothing.
It will be necessary to go on other explorations to be able to add `urls` in theme properties.

PR #297, opened 2022-06-21
fix: remove background images & affiliated urls

lolottetheclash:
This PR reverts https://github.com/okp4/ui/pull/287, https://github.com/okp4/ui/pull/287, https://github.com/okp4/ui/pull/279 and https://github.com/okp4/ui/pull/280

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -40,7 +40,6 @@ export default [
             src: [
               'src/ui/styles/palette.scss',
               'src/ui/styles/themes.scss',
-              'src/ui/styles/urls.scss',
               'src/ui/styles/_exports.module.scss'
             ],
             dest: 'lib/scss'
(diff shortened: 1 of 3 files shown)


fredericvilcot (commented):
@lolottetheclash I guess we also need to remove this [line](https://github.com/okp4/ui/blob/8433f6111ac095abbb9e08e5ac5a3c3ca48a4072/rollup.config.js#L47) from `rollup.config.js`? What do you think?

lolottetheclash:
> @lolottetheclash I guess we also need to remove this [line](https://github.com/okp4/ui/blob/8433f6111ac095abbb9e08e5ac5a3c3ca48a4072/rollup.config.js#L47) from `rollup.config.js`? What do you think?

Thanks for your feedback! I think we can consider going directly through the lib for the images since we plan to be able to thematize the images eventually. What do you think?

fredericvilcot (approved):

PR #325, opened 2022-07-01
fix: remove svg files from rollup-plugin-image scope

fredericvilcot:
This PR prevents `@rollup/plugin-image` to inline svg files instead of `@svgr/rollup`


ccamel (approved):
LGTM!

PR #337, opened 2022-07-05
Fix/highlighted text color

### LINKED ISSUE #333: 🐛 Typography color issues
### **Describe the bug**

The color `highlighted-text` in the `Typography` component is not thematized, making the text unreadable with `light` theme.

Also the documentation for `color` in `Typography` is missing the color `text`

### **To Reproduce**

1. Go to `Typography` in the component [catalog](https://okp4.github.io/ui/?path=/docs/atoms-typography--colors) 
2. Set theme to `light`

### **Expected behavior**

`highlighted-text` is expected to work with both themes

### **Screenshots**

Capture d’écran 2022-07-04 à 15 07 11
Capture d’écran 2022-07-04 à 15 07 21

SonnevilleDamien:
Change the color for highlighted text when changing the theme.

Fix #333

### DIFF
--- a/src/ui/styles/palette.scss
+++ b/src/ui/styles/palette.scss
@@ -47,7 +47,7 @@ $color-light-secondary-selected: #966a13;
 
 // Neutral colors
 $color-light-text: #181c49;
-$color-light-highlighted-text: white;
+$color-light-highlighted-text: #0f224a;
 $color-light-field: #d8d0c6;
 $color-light-border: #ccc2b4;
 $color-light-background: white;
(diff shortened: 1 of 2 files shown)


SonnevilleDamien:
This change of color for the light theme implied a regression, it is usually used on dark background in light theme. The change made the text not readable on other components (ex: languageSwitcher). 

Waiting for a new palette to add a new color.

ErikssonJoakim:
@fredericvilcot @SonnevilleDamien  as the issue have been closed I suggest we change the name of the branch and commit only the `docs` commit, what do you think?

We should also open a new issue in regards to this design problem (style incompatibility)

fredericvilcot:
@ErikssonJoakim 
I close this PR because we must modify before the whole palette with @JulienMassonnat and then fix the story itself

Issue #122, opened 2022-04-08
Select component


ccamel:
## Propose a visual feature

![not-ready](https://img.shields.io/badge/status-notready-yellow.svg)
### **Purpose**

The `Select`component allows the user to choose from a list of options.

### **Visual**

**Dark theme**

<img width="462" alt="Capture d’écran 2022-06-21 à 16 42 07" src="https://user-images.githubusercontent.com/35332974/174827904-50b8bd94-dadd-49b8-9ff6-e212148b9cf1.png">

**Light theme**

<img width="459" alt="Capture d’écran 2022-06-21 à 16 42 15" src="https://user-images.githubusercontent.com/35332974/174827943-89ac181e-090f-4605-a180-91deb58680b4.png">

### **Use cases**

Whenever we want to offer the user the possibility to choose among several options.

#### Typology

`Select` is a molecule component.

#### Properties

`defaultValue?: string`: the value which is displayed to the user when the component is not controlled
`label?: string`: used as the select title
`options: Array<Option>`: a list of objects which contains a label and an associated value
`onChange`: the callback function passed by the parent to specify what is happening when an option is selected
`size?: small | medium | large` : The size of the select, defaultValue will be `medium`
`value`: the value selected by the user. If an empty string is passed `' '`, no option is selected
See API [reference](https://www.radix-ui.com/docs/primitives/components/select#api-reference) 

type Option = {
  label: string,
  value: string,
  group?: string,
}

#### State

- `value?: <Record<string, unknown>`: the selected value

### **Notes**

NA

lolottetheclash:
@ccamel @fredericvilcot @ErikssonJoakim @JoeyDjr Here is the first version of the select component to implement. 
There are certainly additions/modifications to bring, do not hesitate!

ccamel:
Great. Some remarks:
- is it envisaged that the component manages the auto-completion? I don't think so, but I ask.
- don't use `unknown` as type, but a generic `<T>`
- how the defaultValue works? I mean, could you define a default value which is not part of the options? The signature suggests it's the case, but I'd like confirmation.

lolottetheclash:
@ccamel thank you for your feedback.
I think that the `auto-completion` should be implemented in order to propose the most complete component possible, but maybe after the v0. 
The `option`type was not yet defined, it was to be refined later indeed.
I think that the `defaultValue` should be part of the present options, it seems more logical to me but it is possible that I miss a subtlety...
[...]

PR #340, opened 2022-07-06
Feat: Add custom select component

lolottetheclash:
This PR implements Select component whose issue is [here](https://github.com/okp4/ui/issues/122).

As we discuss, the options can be with or without groups:

Capture d’écran 2022-07-29 à 14 01 13

It can also allow multiple choices:

Capture d’écran 2022-07-29 à 14 01 48

We also decided to make the component fully controlled, as mentioned in the story.


JoeyDjr (approved):
Great job ! 👏 🎊

src/ui/molecules/select/Select.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+export type SelectProps = InputPropsForSelect & {
+  /**
+   * Defines the callback called when the select value changes.
+   */
+  readonly onChange: (value: string | Readonly<string[]>) => void
+  /**
+   * The value of the select.
+   */
+  readonly value: string | string[]
```
ErikssonJoakim, line 34: Consider renaming to something like `selected` to make it more clear what the value is

src/ui/molecules/select/Select.tsx
```diff
@@ -0,0 +1,317 @@
...
+  size = 'small',
+  disabled = false,
+  hasError = false,
+  multiple = false,
+  options,
+  onChange,
+  fullWidth,
+  value,
+  helperText
+}: SelectProps): JSX.Element => {
```
ErikssonJoakim, line 71: It's up to you but you could wrap the SelectProps in a deepReadonly and get rid of the Readonly in SelectProps. Makes the type a bit easier to read 😄

ErikssonJoakim (changes requested):
Great work! (In both sense of the word 😄 ). I've left some comments to consider, hope you like them 🍝

src/ui/molecules/select/Select.tsx
lolottetheclash: In general the value selected by the select is called `value`, I prefer to keep this usage.

src/utils.ts
```diff
@@ -24,3 +25,58 @@ export const toPercent = (value: number, min: number, max: number): number => {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const isString = (value: any): value is string => typeof value === 'string'
+
+export const sort = (value1: string, value2: string): number => value1.localeCompare(value2)
```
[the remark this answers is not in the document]
lolottetheclash: Good point! I had forgotten that it was already implemented!

src/utils.ts
```diff
@@ -24,3 +25,58 @@ export const toPercent = (value: number, min: number, max: number): number => {
...
+
+export const compareStrings = (value1: string, value2: string): number =>
```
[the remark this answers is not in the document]
lolottetheclash: According to me the name of the function is clear enough to express its role, a bit like `localeCompare` where `string` does not appear in its name either. 
However, we can keep `refString` and `compareString` to keep consistency with `localeCompare` method documentation. 
This is a good idea to add locales & options in order to offer more flexibility to this utility function 👍

ErikssonJoakim (approved):
Awesome work 🚀

src/utils.ts
```diff
@@ -24,3 +25,62 @@ export const toPercent = (value: number, min: number, max: number): number => {
...
+    return -1
+  }
+  return option2.value.localeCompare(option1.value)
+}
+
+export const getOptionsAscendingSorted = (options: Readonly<Option[]>): Option[] => {
+  return [...options].sort((option1: Option, option2: Option) =>
+    ascendingSortByGroupAndValues(option1, option2)
+  )
+}
```
ccamel, line 72: Apply  [η-conversion](https://wiki.haskell.org/Eta_conversion) (wink to @fredericvilcot) that allows you to drop the abstraction over a function and get rid of this ugly `return`:

```suggestion
export const getOptionsAscendingSorted = (options: Readonly<SelectOption[]>): SelectOption[] =>
  [...options].sort(ascendingSortByGroupAndValues)
```

src/utils.ts
```diff
@@ -24,3 +25,62 @@ export const toPercent = (value: number, min: number, max: number): number => {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const isString = (value: any): value is string => typeof value === 'string'
+
+export const compareStrings = (
```
ccamel, line 29: Please provide tests for all the functions introduced hre.

ccamel (changes requested):

src/utils.ts
```diff
@@ -24,3 +25,62 @@ export const toPercent = (value: number, min: number, max: number): number => {
...
+): number => referenceStr.localeCompare(compareString, locales, options)
+
+const ascendingSortByGroupAndValues = (
+  option1: Readonly<Option>,
+  option2: Readonly<Option>
+): number => {
+  const definedOption1 = option1.group ?? ''
+  const definedOption2 = option2.group ?? ''
+
+  if (definedOption1 > definedOption2) {
```
ccamel, line 45: If you compare the values of the option using `compareStrings`, you should do the same for the groups.

src/utils.ts
```diff
@@ -24,3 +25,62 @@ export const toPercent = (value: number, min: number, max: number): number => {
...
+  locales?: string | string[],
+  // eslint-disable-next-line @typescript-eslint/prefer-readonly-parameter-types
+  options?: Intl.CollatorOptions | undefined
+): number => referenceStr.localeCompare(compareString, locales, options)
+
+const ascendingSortByGroupAndValues = (
+  option1: Readonly<Option>,
+  option2: Readonly<Option>
+): number => {
+  const definedOption1 = option1.group ?? ''
```
ccamel, line 42: Should be named:
```suggestion
  const group1 = option1.group ?? ''
```

src/utils.ts
```diff
@@ -24,3 +25,62 @@ export const toPercent = (value: number, min: number, max: number): number => {
...
+const descendingSortByGroupAndValues = (option1: Option, option2: Option): number => {
+  const definedOption1 = option1.group ?? ''
+  const definedOption2 = option2.group ?? ''
+  if (definedOption1 < definedOption2) {
+    return 1
+  }
+  if (definedOption1 > definedOption2) {
+    return -1
+  }
+  return option2.value.localeCompare(option1.value)
```
ccamel, line 65: Why using a `compareStrings` for ascending, and not the same function for descending?

ccamel:
@lolottetheclash @fredericvilcot I pushed some modifications to the code, some refactoring, and I introduced 2 hooks to address possible transversal concerns. There are still many things to improve such as the use of a memo for the computation of data and we should also answer to the question of whether the component should handle the capitalization of labels. In my opinion no, this should be provided by the caller, in which case we could delete the corresponding code.

ccamel (approved):

JoeyDjr (approved):

fredericvilcot (approved):
@lolottetheclash That's a great piece of work and the component looks sexy, thanks for that! 🚀
We pushed with @ccamel some commits to refactor code, it remains some minor element to refactor that we can discuss later!
Let's merge it!

ccamel (approved):

PR #344, opened 2022-07-07
feat: add progress bar component

JoeyDjr:
This PR is about the Progress bar component as specified in #341.


ccamel:
@fredericvilcot By https://github.com/okp4/ui/pull/344/commits/a66a3528b59c211171030440a8ab67d0fed8e784, I have increased by 1 the limit of the number of compound selectors in `sass` files - if you agree with the process.

FYI @JoeyDjr

JoeyDjr:
> 

Ok thank you. Why not 5 ? Is it too much ?

ccamel:
> Ok thank you. Why not 5 ? Is it too much ?

I don't know it's very opinionated, and it's related to the interpretation of the [SMACSS book](http://smacss.com/book/applicability). For our case, I have just increased the metric for the PR to pass. I guess if we encounter a case where the rule is broken again, well, we'll see what we do.

fredericvilcot:
> @fredericvilcot By [a66a352](https://github.com/okp4/ui/commit/a66a3528b59c211171030440a8ab67d0fed8e784), I have increased by 1 the limit of the number of compound selectors in `sass` files - if you agree with the process.
> 
> FYI @JoeyDjr

@ccamel I fully agree with your approach, let's increase it for now by only 1 and let's see how we could handle selectors in the next future

src/ui/atoms/progressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Primary UI component for progress of a treatment.
+ */
+export const ProgressBar: React.FC<TProgressBarProps> = ({
+  label,
+  minValue = 0,
+  maxValue = 100,
+  currentValue,
+  currentValueFormatter = (currentValue: number): string => `${currentValue}`,
+  progressValueFormatter = (currentValue: number, minValue?: number, maxValue?: number): string =>
```
ccamel, line 60: I think it would be nice to make these functions exported knowing that they can be re-used. I mean:

```typescript
export const defaultValueFormatter = (currentValue: number): string => `${currentValue}`
export const defaultProgressValueFormatter = (currentValue: number, minValue?: number, maxValue?: number): string => `${toPercent(currentValue, minValue ?? 0, maxValue ?? 100).toFixed(2)} %`
```

ccamel (changes requested):
Very small remark. 😌

src/ui/atoms/progressBar/ProgressBar.tsx
JoeyDjr: Ok that's a good idea

fredericvilcot (approved):
Great job!

lolottetheclash (approved):
Nice work, well done !

ErikssonJoakim (approved):
Nice work! 🥇

Issue #375, opened 2022-07-21
🍞 Enhance Toast component


ErikssonJoakim:
### **Purpose**

As of now the content of the `Toast` component are `title` and `description` which are both strings. We want to be able to render any JSX Element in the toaster to give it more flexibility.

### **Visual**

![Capture d’écran 2022-07-21 à 13 45 46](https://user-images.githubusercontent.com/75730728/180206244-309da6ef-44b1-4a2b-b1e5-9dec52e9f1ed.png)

### **Use cases**

A use case with a complex JSX Element should be shown in the `content` story.

#### **Properties**

- description: `JSX.Element | string`

fredericvilcot:
@ErikssonJoakim 
I forgot to tell you, but, `description` property should be an union of both `string` and `JSX.Element` types.
Doing that, it will not bring breaking changes and we could keep string elements in components that already use `Toast`

PR #378, opened 2022-07-21
Feat/improve toast description

ErikssonJoakim:
This PR implements #375

### DIFF
--- a/src/ui/atoms/toast/Toast.tsx
+++ b/src/ui/atoms/toast/Toast.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import * as ToastPrimitive from '@radix-ui/react-toast'
 import { Typography } from 'ui/atoms/typography/Typography'
 import './toast.scss'
+import type { DeepReadonly } from 'superTypes'
 
 type ToastProps = Readonly<{
   /**
@@ -23,7 +24,7 @@ type ToastProps = Readonly<{
   /**
    * An optional description for the toast.
    */
-  readonly description?: string
+  readonly description?: string | JSX.Element
   /**
    * Event handler called when the open state of the dialog changes.
    */
@@ -37,7 +38,7 @@ export const Toast: React.FC<ToastProps> = ({
   title,
   description,
   onOpenChange
-}: ToastProps) => {
+}: DeepReadonly<ToastProps>) => {
   return (
     <ToastPrimitive.Provider swipeDirection="right">
       <ToastPrimitive.Root
@@ -62,9 +63,13 @@ export const Toast: React.FC<ToastProps> = ({
         )}
         {description && (
           <ToastPrimitive.Description asChild className="okp4-toast-description">
-            <Typography as="span" color="highlighted-text" fontSize="small" fontWeight="light">
-              {description}
-            </Typography>
+            {typeof description === 'string' ? (
+              <Typography as="span" color="highlighted-text" fontSize="small" fontWeight="light">
+                {description}
+              </Typography>
+            ) : (
+              description
+            )}
           </ToastPrimitive.Description>
         )}
       </ToastPrimitive.Root>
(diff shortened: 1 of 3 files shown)


stories/atoms/components/toast/Toast.stories.mdx
```diff
@@ -45,7 +47,7 @@ Several severity levels make it possible to fully configure the type of message
 
 <Story
   name="Overview"
-  argTypes={{ isOpened: { control: false } }}
+  argTypes={{ isOpened: { control: false }, description: { control: false } }}
```
ccamel, line 50: It's a disappointment that you can't control that property. Why not force the control as a string to be able to play anyway?

src/ui/atoms/toast/Toast.tsx
```diff
@@ -62,9 +63,13 @@ export const Toast: React.FC<ToastProps> = ({
         )}
         {description && (
           <ToastPrimitive.Description asChild className="okp4-toast-description">
-            <Typography as="span" color="highlighted-text" fontSize="small" fontWeight="light">
-              {description}
-            </Typography>
+            {typeof description === 'string' ? (
```
ccamel, line 66: I'd like to see this condition wrapped into a `typeof` guard function.

stories/atoms/components/toast/Toast.stories.mdx
```diff
@@ -316,6 +321,62 @@ These two elements are optional and are not inter-related.
               severityLevel="success"
             />
           </div>
+          <div>
+            <Button
+              label="Bot Anik"
```
fredericvilcot, line 326: Please, add a better label for the toast button that will be more explicit for the user

stories/atoms/components/toast/Toast.stories.mdx
```diff
@@ -316,6 +321,62 @@ These two elements are optional and are not inter-related.
...
+                  <Typography
+                    as="span"
+                    color="highlighted-text"
+                    fontSize="small"
+                    fontWeight="light"
+                  >
+                    Bot Anik is a bot like no other, she'll kick you around without any trouble! She
+                    gets rid of all who messes around, no one can beat our bot! 🎵
+                  </Typography>
+                  <Typography
```
fredericvilcot, line 343: If you don't mind, I'd prefer the use of the `Button` component instead of a custom element

fredericvilcot (changes requested):

stories/atoms/components/toast/Toast.stories.mdx
ErikssonJoakim: Great catch, I didn't know we could do that with storybook

src/ui/atoms/toast/Toast.tsx
ErikssonJoakim: I've put the guard function in utils as it is a general purpose function

stories/atoms/components/toast/Toast.stories.mdx
ErikssonJoakim: Good point! 👍

fredericvilcot (approved):

ccamel (approved):
👍

JoeyDjr (approved):

lolottetheclash (changes requested):
I think we can remove this part of code: `export const ToastComponent = ({ ... })` because it's not used.

And maybe we can change :
`const handleState = (name, isOpened) => () => {
      setState({ name: name, isOpened: isOpened })
}`
into  :
`const handleState = (name, isOpened) => () => {
      setState({ name, isOpened })
}`

What do you think?

ErikssonJoakim:
@lolottetheclash Great observations! I'd say it's safe to remove those operations

lolottetheclash (approved):
Great !

JoeyDjr (approved):

PR #402, opened 2022-07-28
refactor: rename StateHook type to UseState

JoeyDjr:
This PR is about renaming the type `StateHook` to `UseState` : 
`export type UseState<U> = [U, (value: U) => void]`


ErikssonJoakim (approved):

lolottetheclash (commented):
I think the `UseState` type is too close to the hook exposed by React `useState` and may lead to confusion, what do you think?
Maybe more `State` or `LocalState`? Just a proposal

ccamel (approved):

fredericvilcot (approved):

Issue #413, opened 2022-08-01
List Component


JoeyDjr:
### **Purpose**

The `List` component is a primary UI component that allows display a list of items. Each item is a `ListItem` component.

### **Visual**

![image.png](https://images.zenhubusercontent.com/62b1e6a1f985025ef5e5937e/bd915377-f6e9-4b08-9379-b47fbe73533f)

### **Use cases**

Users can use the `List` to simply display a list of element.

#### **Typology**

The `List` component is an atom.

#### **Properties**

The `List` component is composed of the following properties : 

- items: `ListItem` - The list of the items to be displayed by the `List` 
- leftIcon: `JSX.element` - An icon displayed on the left side
- rightIcon: `JSX.element` - An icon displayed on the right side 

The `ListItem` component is composed of the following properties : 
- name: `string` - The name and main title of the item
- description: `JSX.element` - A content that gives additional information on the item
- leftIcon: `JSX.element` - An icon displayed on the left side
- rightIcon: `JSX.element` - An icon displayed on the right side 

#### **State**

N/A

### **Notes**

N/A

PR #415, opened 2022-08-02
List component

JoeyDjr:
This PR is about the implementation of the list component and the docs (storybook) of this component.
👉 Issue related : https://github.com/okp4/ui/issues/413


ccamel (approved):
Nice. 👍

src/ui/atoms/list/List.tsx
```diff
@@ -0,0 +1,43 @@
...
+   */
+  readonly items: ListItemProps[]
+  /**
+   * An icon displayed on the left side of each items.
+   */
+  readonly leftIcon?: JSX.Element
+  /**
+   * An icon displayed on the right side of each items.
+   */
+  readonly rightIcon?: JSX.Element
```
fredericvilcot, line 19: I can't see clearly here the purpose of the right and left icons 🤔
IMHO, I think these properties should be only the responsibility of the `ListItem` because the `List` only maps and renders child items, it shouldn't influence the content of an element.
Furthermore, it seems to be redundant with `ListItem` properties.

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+   */
+  readonly description?: string | JSX.Element
+  /**
+   * An icon that can provide visual information on the left side.
+   */
+  readonly leftIcon?: JSX.Element
+  /**
+   * An icon that can provide visual information on the right side.
+   */
+  readonly rightIcon?: JSX.Element
```
fredericvilcot, line 23: As I said above, I am not sure about those icon's properties, for many reasons:
- it could be something other than icons (text for example, a button, etc.)
- in my opinion, it should be part of the description itself (or the main content of the `ListItem`)
- the style is imposed, maybe that should be the responsibility of the caller? (it's an open question)

stories/atoms/components/list/List.stories.mdx
```diff
@@ -0,0 +1,284 @@
...
+    },
+    items: {
+      table: {
+        disable: true
+      }
+    }
+  }}
+  args={{
+    label: 'Item',
+    description: 'Description',
```
fredericvilcot, line 135: I would have liked to have the control of the label and the description in the table of arguments, but a priori it does not work...
image

stories/atoms/components/list/List.stories.mdx
```diff
@@ -0,0 +1,284 @@
...
+        ]}
+      />
+    )}
+  </Story>
+</Canvas>
+
+## Icons
+
+The `leftIcon` and `rightIcon` properties allow you to provide visual information for each item in the list.  
+Theses properties override the `leftIcon` and `rightIcon` properties of the `List` if they are provided.
```
fredericvilcot, line 191: This section is quite confusing, related to my comments in components on icon properties

fredericvilcot (changes requested):
The final rendering is great, well done! 🚀
I have noted a few remarks on which we can discuss together 😌

stories/atoms/components/list/List.stories.mdx
JoeyDjr: Yes it doesn't work because it is the properties table of the `ListItem` component (not the `List`). It is generated by `<ArgsTable of={ListItem} />` and not some thing like : `<ArgsTable story="StoryName" />`
So we can't control the properties here.

fredericvilcot: @JoeyDjr 
Thanks for your explanation, I understand better now 👌
According to what you said, I'll suggest to write a dedicated story for the `ListItem` component and make the link between the two stories.

stories/atoms/components/list/List.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+## Description
+
+The `List` allows to display a list of elements and manage them.
+
+## Overview
+
+<Story
+  name="Overview"
+  argTypes={{
+    items: { control: false }
```
fredericvilcot, line 58: Could you enhance the `items` type in the story?
The arguments' table shows `ListItemProps[]` as type but I'm not sure this is understandable for the user..

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * The main title of the item.
+   */
+  readonly title?: string
+  /**
+   * Additional information about the item
+   */
+  readonly description?: string | JSX.Element
+  /**
+   * An that can provide visual information or allow interactions on the left side.
```
fredericvilcot, line 17: I think you are missing a word here ;)
(and also for the leftElement comment)

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <div className="okp4-listitem-main">
+      {leftElement && <div className="okp4-listitem-left-element">{leftElement}</div>}
+      {title && (
+        <div className="okp4-listitem-title">
+          <Typography as="div" color="text" fontSize="small" fontWeight="bold">
+            {title}
+          </Typography>
+        </div>
+      )}
+      {description && <div className="okp4-listitem-description">{renderDescription()}</div>}
```
fredericvilcot, line 51: I think you can move assertion and the `div` container directly in the `renderDescription` function to be more readable, what do you think?

stories/atoms/components/list/List.stories.mdx
```diff
@@ -0,0 +1,276 @@
...
+  }}
+>
+  {args => <List {...args} />}
+</Story>
+
+## Properties
+
+<ArgsTable story="Overview" />
+
+## ListItem
```
fredericvilcot, line 102: Please make a dedicated story for the `ListItem`

fredericvilcot (changes requested):
@JoeyDjr 
Thanks for the changes, I think it's really much better 🚀
I have some minor remarks that I think we can easily handle before merging

stories/atoms/components/list/List.stories.mdx
JoeyDjr: Ok. Is it ok to put `ListItem` which is defined right after?

fredericvilcot: I think we should expose the real type, so the user does not have to switch on `ListItem` story to look for it.
```suggestion
    items: { 
      control: false, 
      table: {
        type: { 
          summary: 'Array<{title?: string, description?: string | JSX.Element, leftElement?: JSX.Element, rightElement?: JSX.Element}>' }
      } }
```

fredericvilcot (approved):
LGTM! 👌

ccamel (approved):

PR #461, opened 2022-08-23
Feat: add getFiles selector

lolottetheclash:
The purpose of this PR is to add a selector in order to recover all the loaded files

### DIFF
--- a/src/domain/file/store/selector/file.selector.spec.ts
+++ b/src/domain/file/store/selector/file.selector.spec.ts
@@ -1,0 +1,63 @@
+import { OrderedMap, OrderedSet } from 'immutable'
+import { EventBus } from 'ts-bus'
+import type { DeepReadonly } from 'superTypes'
+import type { AppState } from '../../store/appState'
+import { FileStoreBuilder } from '../builder/store.builder'
+import { FileBuilder } from '../../builder/file.builder'
+import { FileDescriptor, getFiles } from './file.selector'
+import { FileEntity } from 'domain/file/entity/file'
+
+type Data = DeepReadonly<{
+  state: AppState
+  expectedFiles: ReturnType<typeof getFiles>
+}>
+
+const file1 = new FileBuilder()
+  .withId('id1')
+  .withName('image1')
+  .withSize(100)
+  .withType('image/png')
+  .withStream(new ReadableStream())
+  .build()
+
+const file2 = new FileBuilder()
+  .withId('id2')
+  .withName('image2')
+  .withSize(100)
+  .withType('text/csv')
+  .withStream(new ReadableStream())
+  .build()
+
+const preloadedState: AppState = {
+  file: {
+    byId: OrderedMap<string, FileEntity>().set(file1.id, file1).set(file2.id, file2),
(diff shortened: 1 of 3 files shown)


JoeyDjr (approved):

ErikssonJoakim (approved):
Cool stuff 👍

src/domain/file/store/selector/file.selector.ts
```diff
@@ -0,0 +1,18 @@
+import { createSelector } from 'reselect'
+import type { DeepReadonly } from 'superTypes'
+import type { AppState, FileById } from '../appState'
+import type { MediaType } from 'domain/common/type'
+import type { FileEntity } from 'domain/file/entity/file'
+
+export type FileDescriptor = {
```
fredericvilcot, line 7: We also need the `id` property here that must be mandatory.

src/domain/file/store/selector/file.selector.ts
```diff
@@ -0,0 +1,18 @@
...
+  readonly name: string
+  readonly size: number
+  readonly type: MediaType
+}
+
+export const getFiles: (state: DeepReadonly<AppState>) => FileDescriptor[] = createSelector(
+  (state: DeepReadonly<AppState>): FileById => state.file.byId,
+  (files: DeepReadonly<FileById>): FileDescriptor[] => [
+    ...files.map(({ name, size, type }: FileEntity) => ({ name, size, type })).toList()
```
fredericvilcot, line 16: I think here we can avoid the use of the `immutable` `map` function that returns a new `OrderedMap` which weighs down the function for nothing. 
IMHO, It's better to first convert the `OrderedMap` into an JS `Array` and then `map` on it.
```suggestion
    files
      .toIndexedSeq()
      .toArray()
      .map(({ id, name, size, type }: FileEntity) => ({ id, name, size, type }))
```

src/domain/file/store/selector/file.selector.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  file: {
+    byId: OrderedMap<string, FileEntity>().set(file1.id, file1).set(file2.id, file2),
+    byType: OrderedMap<string, OrderedSet<string>>()
+      .set(file1.type, OrderedSet<string>([file1.id]))
+      .set(file2.type, OrderedSet<string>([file2.id]))
+  }
+}
+
+const expectedFiles = (): FileDescriptor[] =>
+  [file1, file2].map(({ name, size, type }) => ({ name, size, type }))
```
fredericvilcot, line 41: We need the `id` property in the `expectedFiles`

fredericvilcot (changes requested):
Hey, your first ever PR on domain business, great! 
I wrote down some misses, let me know if you need more explanations!

lolottetheclash:
> Hey, your first ever PR on domain business, great! I wrote down some misses, let me know if you need more explanations!

Thanks for all your relevant remarks 👌I'll fix it right away

ErikssonJoakim (approved):

fredericvilcot (approved):
Great!

amimart (approved):
Seems good!

JoeyDjr (approved):

PR #462, opened 2022-08-23
refactor(file input): rename props type

JoeyDjr:
This PR rename `FileUploadProps` to `FileInputProps` to be consistent with the component name (`File Input`)


lolottetheclash (approved):
good catch !

fredericvilcot (approved):

Issue #459, opened 2022-08-23
📏 Refactor component stories


ErikssonJoakim:
📄 

Now that we have a dedicated wiki for [how to create a story](https://github.com/okp4/ui/wiki/Create-A-Story) we should refactor our existing component stories so that they follow these requirements.
**ATOMS**
- [ ] #623
- [x] #466
- [x] #624
- [x] #543 
- [ ] #625
- [x] #486 
- [x] #626
- [ ] #627
- [ ] #628
- [ ] #629
- [x] #630
- [ ] #631
- [ ] #632
- [x] #633

**MOLECULES**
- [ ] #634
- [ ] #635
- [ ] #636
- [ ] #637

**ORGANISMS**
- [ ] #638

MdechampG:
@ErikssonJoakim : to split

fredericvilcot:
Will be part of new incoming components

PR #463, opened 2022-08-24
Refactor/refactor babylon engine story

ErikssonJoakim:
This PR implements #459 for `BabylonEngine`

For the moment I have manually changed the type summary of the options property but it would be preferable to have this done automatically if possible.

Keeping this PR as a draft as it needs some more work, most importantly stories for it's properties.


ErikssonJoakim:
Temporary closing PR for later

PR #467, opened 2022-08-25
fix: move lodash from devDependencies to dependencies

fredericvilcot:
This PR fixes the way `lodash` is declared in the lib by moving the package from `devDependencies` to `dependencies`.
This ensures that the package will be automatically installed into apps that consumes `@okp4/ui`


lolottetheclash (approved):
👌

JoeyDjr (approved):

ErikssonJoakim (approved):

amimart (approved):

PR #476, opened 2022-08-26
Refactor/task domain

fredericvilcot:
👉 This PR modifies the way commands are handled in the `task` domain.

😱 Indeed, in the previous version, the fundamental principle of protection of domain entities under hexagonal architecture was violated because the commands called by the primary controller (listener) directly manipulated the `task` or the `updateTask` entities as a parameter.

✅ For this, it is necessary to create a type of transfer dissociated from the entity as well as to add a mapping layer which will ensure that the entity is correctly constructed from the DTO.

😎 The `task` entity is thus not leaked to the controllers and the events only transport DTOs.

### DIFF
--- a/src/adapters/task/mapper/task.mapper.ts
+++ b/src/adapters/task/mapper/task.mapper.ts
@@ -1,7 +1,8 @@
 /* eslint-disable functional/no-let */
 import { TaskBuilder } from 'domain/task/builder/task/task.builder'
 import type { Task, UpdateTask } from 'domain/task/entity/task'
-import type { AmendTask, CreateTask } from 'domain/task/command/createTask'
+import type { AmendTask } from 'domain/task/command/amendtask'
+import type { CreateTask } from 'domain/task/command/createTask'
 import type { DeepReadonly } from 'superTypes'
 import { UpdateTaskBuilder } from 'domain/task/builder/updateTask/updateTask.builder'
 
--- a/src/adapters/task/mapper/task.mapper.ts
+++ b/src/adapters/task/mapper/task.mapper.ts
@@ -1,0 +1,17 @@
+import { TaskBuilder } from 'domain/task/builder/task/task.builder'
+import type { Task } from 'domain/task/entity/task'
+import type { CreateTask } from 'domain/task/command/createTask'
+import type { DeepReadonly } from 'superTypes'
+
+export class TaskMapper {
+  public static readonly mapCommandPayloadToEntity = (payload: DeepReadonly<CreateTask>): Task =>
+    new TaskBuilder()
+      .withId(payload.id)
+      .withCreationDate(payload.timestamp)
+      .withLastUpdateDate(payload.timestamp)
+      .withInitiator(payload.initiator)
+      .withMessageKey(payload.messageKey)
+      .withType(payload.type)
+      .withStatus(payload.status)
+      .build()
+}
--- a/src/adapters/task/mapper/task.mapper.ts
+++ b/src/adapters/task/mapper/task.mapper.ts
@@ -1,10 +1,11 @@
 import { TaskBuilder } from 'domain/task/builder/task/task.builder'
-import type { Task } from 'domain/task/entity/task'
-import type { CreateTask } from 'domain/task/command/createTask'
+import type { Task, UpdateTask } from 'domain/task/entity/task'
+import type { AmendTask, CreateTask } from 'domain/task/command/createTask'
 import type { DeepReadonly } from 'superTypes'
+import { UpdateTaskBuilder } from 'domain/task/builder/updateTask/updateTask.builder'
 
 export class TaskMapper {
-  public static readonly mapCommandPayloadToEntity = (payload: DeepReadonly<CreateTask>): Task =>
+  public static readonly mapCreateTaskToEntity = (payload: DeepReadonly<CreateTask>): Task =>
     new TaskBuilder()
       .withId(payload.id)
       .withCreationDate(payload.timestamp)
@@ -14,4 +15,12 @@ export class TaskMapper {
       .withType(payload.type)
       .withStatus(payload.status)
       .build()
+
+  public static readonly mapAmendTaskToEntity = (payload: DeepReadonly<AmendTask>): UpdateTask => {
+    // eslint-disable-next-line functional/no-let
+    let builder = new UpdateTaskBuilder().withId(payload.id).withLastUpdateDate(payload.timestamp)
+    if (payload.messageKey) builder = builder.withMessageKey(payload.messageKey)
+    if (payload.status) builder = builder.withStatus(payload.status)
+    return builder.build()
+  }
 }
--- a/src/adapters/task/mapper/task.mapper.ts
+++ b/src/adapters/task/mapper/task.mapper.ts
@@ -1,3 +1,4 @@
+/* eslint-disable functional/no-let */
 import { TaskBuilder } from 'domain/task/builder/task/task.builder'
 import type { Task, UpdateTask } from 'domain/task/entity/task'
 import type { AmendTask, CreateTask } from 'domain/task/command/createTask'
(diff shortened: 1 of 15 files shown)


src/domain/task/command/createTask.ts
```diff
@@ -0,0 +1,17 @@
...
+export type CreateTask<I = string, T = string> = {
+  readonly id: I
+  readonly timestamp: Date
+  readonly type: T
+  readonly status: TaskStatus
+  readonly messageKey: string
+  readonly initiator?: string
+}
+
+export type AmendTask<I = string> = {
```
amimart, line 12: Maybe we can put it in another file ? Or rename this one because it doesn't seems to fit very well in a `createTask.ts` file

amimart (changes requested):
Seems almost good to me 😉

src/domain/task/command/createTask.ts
fredericvilcot: Good catch, 100% agree 😌

amimart (approved):

src/domain/task/usecase/register-task/registerTask.ts
```diff
@@ -1,17 +1,19 @@
...
+import type { DeepReadonly } from 'superTypes'
 
 const dispatchError = (error: unknown, dispatch: ReduxStore['dispatch']): void => {
   const errorToDispatch = ErrorMapper.mapRawErrorToEntity(error)
   dispatch(ThrowErrorActions.errorThrown(errorToDispatch))
 }
 
 export const registerTask =
-  (task: Task): ThunkResult<Promise<void>> =>
+  (task: DeepReadonly<CreateTask>): ThunkResult<Promise<void>> =>
```
ccamel, line 16: If the UC is registerTask, I would expect the command to be named `RegisterTask`.

src/domain/task/usecase/register-task/registerTask.ts
```diff
@@ -23,5 +25,5 @@ export const registerTask =
       )
       return
     }
-    dispatch(RegisterTaskActions.taskRegistered(task))
+    dispatch(RegisterTaskActions.taskRegistered(TaskMapper.mapCreateTaskToEntity(task)))
```
ccamel, line 28: 🤔 As far as I understand, the entity task leaks in the event. It should not.

src/adapters/task/mapper/task.mapper.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable functional/no-let */
+import { TaskBuilder } from 'domain/task/builder/task/task.builder'
+import type { Task, UpdateTask } from 'domain/task/entity/task'
+import type { AmendTask } from 'domain/task/command/amendtask'
+import type { CreateTask } from 'domain/task/command/createTask'
+import type { DeepReadonly } from 'superTypes'
+import { UpdateTaskBuilder } from 'domain/task/builder/updateTask/updateTask.builder'
+
+export class TaskMapper {
+  public static readonly mapCreateTaskToEntity = (payload: DeepReadonly<CreateTask>): Task => {
```
ccamel, line 10: IMHO, there should be no mapping between the command and the entity, since it is the role of the CPU to manage it. So I would advise to delete it.

src/adapters/task/mapper/task.mapper.ts
```diff
@@ -0,0 +1,28 @@
...
+      .withCreationDate(payload.timestamp)
+      .withLastUpdateDate(payload.timestamp)
+      .withMessageKey(payload.messageKey)
+      .withType(payload.type)
+      .withStatus(payload.status)
+    if (payload.initiator) builder = builder.withInitiator(payload.initiator)
+    return builder.build()
+  }
+
+  public static readonly mapAmendTaskToEntity = (payload: DeepReadonly<AmendTask>): UpdateTask => {
```
ccamel, line 22: I hadn't noticed it before, but I find it strange that `UpdateTask` is an entity. There should be only one entity in the domain: `Task`.

ccamel (changes requested):
A quick review with some remarks. As far as I see, there are issues in the definition of types, from a CQRS-ES point of view between commands, entity and events (payloads in particular). This should be taken care of because it is quite structuring.

Let me know if want more information.

src/adapters/task/mapper/task.mapper.ts
ccamel: To be clearer:

**entities**:
- Task

**commands**:
- registerTask
- amendTask

events:
- TaskRegistered + TaskRegisteredPayload
- TaskAmended + TaskAmendedPayload

src/domain/task/usecase/register-task/registerTask.ts
fredericvilcot: Yes, I know.
But, there is no option here, we need to call a redux action which calls the reducer that updates the state.
For that, we are forced to give the entity as the payload. If not, this will be the reducer itself to build the entity and I don't think this is a good way to achieve that.
The issue here is that the event bus is coupled with redux as a middleware.
So, if we want to avoid the entity leak in the event bus, we need to find a way to 'deserialize' the payload before been published by the bus.
I think it is a deep change that should be addressed in a specific PR, what do you think?

ccamel: @fredericvilcot I don't understand the point. As far as I see, the dispatch + reducer is compatible with ES requirements. You can dispatch an event (which is not an entity) and in the reducer, given what is provided in the event recombine the aggregate. I don't see why you can't.

fredericvilcot: We never handled any business logic in reducers in the previous domains, so payloads are the entities themselves...
To modify that, we need a bigger refactor.

ccamel: Well it's not business logic. The business logic remains in the UC. The reducer just applies (in a pure fashion) the events in the store. So it's really not a big deal, and I think it's a refactor to do here, scoped on this domain (because it's rather simple). It's important to do it right because otherwise you won't be able to scale (and sadly we will continue to break the fundamental rules of the Hexagonal and the Event Sourcing).

fredericvilcot: @ccamel 
Ok, so, if I understand, the UC is simply here a flat pass that receives command payload as input and only dispatches a redux action with the same payload? 
It will be the reducer to gets this payload and build this entity?

src/adapters/task/mapper/task.mapper.ts
fredericvilcot: OK, the builder is directly called in the reducer itself?

src/domain/task/usecase/register-task/registerTask.ts
ccamel: @fredericvilcot It's a little more subtle. The UC is not a pass through: it is responsible of the logic for validating the command, ensure it is acceptable regarding the current state and finally generates the *different* events. Here's is the business logic: translate a command into several events. And events should be DTOs only (and immutable of course).

The reason why the entity should not be created in the UC is that the different events published can be used to build different models (different Read models). The entity here is only used for the Write model.

Then, in the reducer, we're on the write side, you just apply the events to have the store updated. There's nothing wrong in handling the entities here. It's what a reducer is supposed to do.

Does it make more sense? It's quite difficult to explain in a few words.

src/adapters/task/mapper/task.mapper.ts
ccamel: Not if the builder has business logic, because if it is the case, it causes a problem of responsibility.

fredericvilcot:
After relevant discussions with @ccamel, it appears that:

- the hexagonal architecture is not violated anymore as entities remain in domain
- BUT the ES / CQRS principles are violated as the use-cases dispatch a redux event with entities as payload and by the way
- those events can be consumed by controller that subscribe to the event bus (leak)
- to be more compliant with ES and CQRS, all commands and events must take transition objects as payloads witch will be handled later to build / update entities.

For that, I put this PR to draft to achieve this modifications.

src/domain/task/usecase/register-task/registerTask.ts
fredericvilcot: @ccamel 
I agree with that approach but it remains one interrogation: what about handling possible error when building the entity in the reducer? What about if the entity must load more external context to ne build? 
The reducer is a pure function so we can't apply here side effects like throwing or calling methods that should fail.
So, how can we handle this part of the write side?

ccamel: @fredericvilcot You can't have errors in the reducer. If the event has been published by the UC, then it can't be denied in any case, it must be applied and produce a change (a transition) in the state.

What do you mean by "load more external context"? In the reducer, you can't access other resources than the store (it's a pure function). If you need more to build the entity, it has to be carried by the event and resolved by the UC which can access all the secondary adapters it wants.

fredericvilcot: @ccamel 
Ok, thanks for your explanations.

To respect the ES principles, I will move invariants from builders to the UC to ensure that the emitted event is foolproof.

So, as a result, the reducer will only have to worry about aggregate's recomposition by managing the state.

Therefore, IMHO, entity builders can no longer be used directly in the UC except for testing because the invariants must be checked before the event is propagated and therefore before the entity itself is built.

If you agree with that, I can resolve this comment and implement the changes.

Issue #468, opened 2022-08-25
File Picker


JoeyDjr:
### **Purpose**

The `FilePicker` component is a component that displays a file input and the list of the selected files. It is a combination of the `FileInput` component and the `List` and `ListItem` components.

### **Visual**

![image.png](https://images.zenhubusercontent.com/62b1e6a1f985025ef5e5937e/bd36b784-81dd-4b21-988d-ab417d26023c)

### **Use cases**

Users can use the `FilePicker` to selecte files and view the selected files.

### **Specification**

This component is able to check the files extensions according to the provided property and display an error if at least one file does not match.

The component is provided with i18n files to display the error message and translate the file size unit of the file (MB, Mo, etc.).

#### **Typology**

The `FilePicker` component is an organism. It dispatch action to stores the files when they are dropped, and to get the files to display them in the list. When a error occured, it sends an error.

#### **Properties**

The `FilePicker` component is composed of the following properties :

- First of all, it has the following properties of `FileInput` : `label`, `description`, `multiple`, `acceptedFormats`, `size`.

#### **State**

The `FilePicker` can be into different state : 
- Display only the file input. If no file is selected
- Display of the file input and the list of selected files.
- One of the previous states and the file input in an error state with an error message.

The description of the list item is the file size. A function will automatically display the readable size. For example : "1,1 Mo", "4,6 Go", "154 KB", "875 octets", ...

### **Notes**

For the v0, the error are handled in the component. But later, in a future version, we will be able to implement that errors and error messages are more customizable.

👉 This issue refers to this discussion : https://github.com/okp4/ui/discussions/464

PR #488, opened 2022-08-30
File Picker component

JoeyDjr:
This PR is about the `File Picker` component.

👉 Issue  https://github.com/okp4/ui/issues/468

image

The File Picker is a combination of `FileInput` and `List ` of the selected files.

The PR adds/fixes a feature about adding same file in the `FileInput` too.


src/utils.ts
```diff
@@ -61,3 +61,59 @@ export const sortSelectOptionAsc = (options: Readonly<SelectOption[]>): SelectOp
...
+ * @param files The files to check.
+ * @param acceptedFormats The accepted formats (MIME type or extension) of the files.
+ * @returns A boolean that indicates if the files match the MIME types or extensions.
+ */
+export const areFilesAccepted = (
+  files: DeepReadonly<File[]>,
+  acceptedFormats?: Readonly<string[]>
+): boolean =>
+  files.every((file: DeepReadonly<File>) => {
+    return (
```
fredericvilcot, line 114: Please, remove the `return` keyword to a more concise ES6 syntax.

src/ui/organisms/filePicker/FilePicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Displays a clickable text to delete all selected files.
+   */
+  readonly clearAll?: boolean
+}
+
+// eslint-disable-next-line max-lines-per-function
+export const FilePicker: React.FC<FilePickerProps> = ({
+  clearAll,
+  acceptedFormats,
```
fredericvilcot, line 33: Redundant, it is part of the spreaded props from `FileInput`
```suggestion
```

src/ui/atoms/fileInput/FileInput.tsx
```diff
@@ -8,35 +8,35 @@ import { truthy } from 'utils'
...
    */
   readonly multiple?: boolean
   /**
-   * The list of the accepted formats
+   * The list of the accepted formats.
    */
   readonly acceptedFormats?: string[]
   /**
-   * The size of the component
+   * The size of the component.
```
fredericvilcot, line 27: I would have added more precision about the size to add more context to the `FilePicker` usage
```suggestion
   * The size of the file input area.
```

src/ui/atoms/fileInput/FileInput.tsx
```diff
@@ -75,6 +75,15 @@ export const FileInput: React.FC<FileInputProps> = ({
     event.stopPropagation()
   }, [])
 
+  const allowSelectSameFile = useCallback(
```
fredericvilcot, line 78: IMHO, the naming should be more explicit
```suggestion
  const clearInputValue = useCallback(
```

fredericvilcot (changes requested):
That's a great work 💪
I wrote down some remarks, let me know 😌

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -0,0 +1,164 @@
...
+const fileStoreParameter: StoreParameter = [FileContext, fileStore]
+
+export default List([fileStoreParameter])
+```
+
+### 2. Wrap the `FilePicker` component into a store provider
+
+Finally, it is necessary to wrap the `FilePicker` component in a `StoreProvider` which will auto-inject into the React context the configuration elements necessary for its proper functioning.
+
+This `StoreProvider` component is directly provided by the `@okp4/ui` library, you simply have to pass it the parameters exported by the `store.ts` file as properties.
```
ErikssonJoakim, line 148:
```suggestion
### 2. Wrap the FilePicker component into a store provider

Finally, it is necessary to wrap the FilePicker in a StoreProvider which will auto-inject into the React context the configuration elements necessary for its proper functioning.

This StoreProvider is directly provided by the `@okp4/ui` library, you simply have to pass it the parameters exported by the **store.ts** file as properties.
```

ErikssonJoakim (changes requested):
I got some remarks, mostly about the story, that we can discuss.  Other than that great work! 🚀 

Also could you please change the default value of `multiple` to **false** in fileInput.tsx? It makes more sense for booleans to have **false** as default.

fredericvilcot:
@JoeyDjr 
Could please update the screenshots in the main PR description? 🙏

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -0,0 +1,164 @@
...
+  title="Organisms/FilePicker"
+  component={FilePicker}
+  parameters={{
+    actions: { argTypesRegex: '^on.*' },
+    docs: {
+      source: {
+        type: 'code'
+      }
+    }
+  }}
```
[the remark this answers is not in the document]
JoeyDjr: ok 👍

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -0,0 +1,164 @@
...
+# OKP4 File Picker
+
+> Allows to select files and display them.
+
+[![stability-unstable](https://img.shields.io/badge/stability-unstable-yellow.svg)](https://github.com/emersion/stability-badges#unstable)
+
+## Description
+
+The `FilePicker` component is a component that displays a file input and the list of the selected files.  
+It is a combination of the [FileInput](/docs/atoms-fileinput--overview) component and the [List](/docs/atoms-list--overview) and [Listitem](/docs/atoms-listitem--overview) components.
```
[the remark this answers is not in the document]
JoeyDjr: I agree to rewrite the description. But I think that the word "upload" is not appropriate here because the component does not allow to upload the files. 
What do you think @lolottetheclash @ccamel @fredericvilcot ?

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -0,0 +1,164 @@
...
+      <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'center' }}>
+        <Typography as="div" fontFamily="brand" fontSize="small" fontWeight="xlight">
+          Supports images like .png, .jpg, .jpeg, etc. and CSV files
+        </Typography>
+      </div>
+    ),
+    multiple: true,
+    acceptedFormats: ['image/*', '.csv'],
+    size: 'large',
+    clearAll: true
```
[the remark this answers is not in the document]
JoeyDjr: Ok thanks
Ok for `size`
But for `clearAll` it is an error in the component because we want this property to have `true` by default

ErikssonJoakim (approved):
Great work 💪

src/ui/organisms/filePicker/FilePicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+    (id: string) => (): ThunkResult<Promise<void>> => fileDispatch(removeFile(id)),
+    [fileDispatch]
+  )
+
+  const handleRemoveAll = useCallback(
+    (): ThunkResult<Promise<void>> => fileDispatch(removeAllFiles()),
+    [fileDispatch]
+  )
+
+  const RemoveIcon = (id: string): JSX.Element => (
```
lolottetheclash, line 83: Concerning the naming, it seemed to me that we said that we write in PascalCase when the component is then called like this `<MyComponent id={id1} />` and that we use the camelCase when we call it like this: `myComponent(id1)`, or I am wrong?
I have the impression that here it is a mix of both.

src/ui/organisms/filePicker/FilePicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+                size: file.size,
+                type: file.type,
+                stream: file.stream() as unknown as ReadableStream
+              }
+            })
+          )
+        )
+      } else {
+        setError(true)
+        setErrorMessage(t(`filePicker:filePicker.errorMessage.type`))
```
lolottetheclash, line 67: You don't need back ticks here.

src/ui/organisms/filePicker/FilePicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+        {...props}
+        acceptedFormats={asMutable(props.acceptedFormats)}
+        error={isError}
+        errorMessage={errorMessage}
+        onDropped={handleDropped}
+      />
+      {clearAll && fileList.length > 1 && (
+        <div className="okp4-file-picker-clear-all" onClick={handleRemoveAll}>
+          <Typography fontSize="x-small" fontWeight="xlight" textDecoration="underline">
+            {t(`filePicker:filePicker.clearAll`)}
```
lolottetheclash, line 114: You don't need back ticks here.

lolottetheclash (changes requested):
This is a great work! Well done! 💪
I have a thought about the display of the list of selected files: when you add some as you go along, the most recently loaded is displayed at the end of the list.
I would have tended to do the opposite, i.e. display the most recent one at the top of the list because if you realize that it was not the one you wanted to load, it is immediately accessible to delete it (instead of scrolling if there are many files). 
What do you think about it?

src/ui/organisms/filePicker/i18n/filePicker_fr.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filePicker": {
+    "fileSizeUnit": {
+      "B": "octets",
```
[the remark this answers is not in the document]
JoeyDjr: I think it's better to put the `bytes` in lower case.

src/ui/organisms/filePicker/FilePicker.tsx
```diff
@@ -0,0 +1,121 @@
...
+    return `${value} ${t(`filePicker:filePicker.fileSizeUnit.${unit}`)}`
+  }
+
+  const handleDropped = useCallback(
+    (files: DeepReadonly<File[]>) => {
+      if (areFilesAccepted(files, props.acceptedFormats)) {
+        setError(false)
+        fileDispatch(
+          storeFiles(
+            files.map((file: DeepReadonly<File>) => {
```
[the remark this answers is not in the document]
JoeyDjr: I agree. It's a good idea

src/ui/organisms/filePicker/i18n/filePicker_en.json
```diff
@@ -0,0 +1,15 @@
+{
+  "filePicker": {
+    "fileSizeUnit": {
+      "B": "Bytes",
+      "KB": "KB",
```
ccamel, line 5: The metric prefix for `kilo` is [`k`](https://en.wikipedia.org/wiki/Metric_prefix) - (see [kilobyte](https://en.wikipedia.org/wiki/Kilobyte)).

```suggestion
      "KB": "kB",
```

src/ui/organisms/filePicker/i18n/filePicker_fr.json
```diff
@@ -0,0 +1,15 @@
...
+      "KB": "Ko",
```
ccamel, line 5:
```suggestion
      "KB": "ko",
```

src/ui/organisms/filePicker/FilePicker.tsx
JoeyDjr: Yes maybe it is better to write : 
```tsx
const RemoveIcon = ({ id }: Readonly<{ id: string }>): JSX.Element => (
    <div className="okp4-file-picker-list-item-delete" onClick={handleRemove(id)}>
      <Icon invertColor name="close" />
    </div>
  )
  
  ...
  
  rightElement={<RemoveIcon id={id} />}
  ```
  
  what do you think @lolottetheclash ?

Comment (lolottetheclash) [reply src/ui/organisms/filePicker/FilePicker.tsx]: Yeah, I think it's better too. 
But I wonder if we should make a type for this: `Readonly<{ id: string }>` to become this:
`Readonly<RemoveIconProps>`

`type RemoveIconProps = {
  id : string
}`

or if it's ok to leave it like that... @ccamel @fredericvilcot ?

Comment (ccamel) [reply src/ui/organisms/filePicker/FilePicker.tsx]: For an internal component, I would leave it like that.

Comment (lolottetheclash) [reply src/ui/organisms/filePicker/FilePicker.tsx]: Good to know ! thanks ;)

**Comment by JoeyDjr** [comment]
> This is a great work! Well done! 💪 I have a thought about the display of the list of selected files: when you add some as you go along, the most recently loaded is displayed at the end of the list. I would have tended to do the opposite, i.e. display the most recent one at the top of the list because if you realize that it was not the one you wanted to load, it is immediately accessible to delete it (instead of scrolling if there are many files). What do you think about it?

I think it's a good point.

**Review by ccamel — APPROVED** [review]

File: src/utils.ts [hunk 16]
```diff
@@ -1,4 +1,6 @@
-import type { DeepReadonly } from './superTypes'
+import type { DeepReadonly, FileSizeUnit, ReadableFileSize } from './superTypes'
+import short from 'short-uuid'
+import type { StoreFilesPayload } from 'domain/file/command/storeFile'
```
fredericvilcot, line 3: The `utils` file can't be coupled with the domain itself, you should depend on abstractions.

src/utils.ts
```diff
@@ -61,3 +63,76 @@ export const sortSelectOptionAsc = (options: Readonly<SelectOption[]>): SelectOp
...
+      !acceptedFormats.length ||
+      acceptedFormats.some((format: string) => isFileTypeAccepted(file, format))
+  )
+
+/**
+ * Convert a list of `File` into a list of `StoreFilesPayload` to store in the redux store.
+ * @param files The list of file to convert
+ * @returns A list of converted files.
+ */
+export const toStoreFilesPayload = (
```
fredericvilcot, line 127: IMHO, this method should not be part of `utils` as it exposes the way the file is mapped into a command payload, this is the domain responsibility.
I would have moved the whole method into the controller itself and thus, we could implement later a `builder` in the `File` domain that could do this job.

ErikssonJoakim (approved):

fredericvilcot (changes requested):
Great job, we are almost here! 💪
I wrote down some remarks that should be handled as they are structurant, let me know if you have some interrogations 😌

fredericvilcot (approved):
LGTM!

lolottetheclash (approved):
Well done !

PR #491, opened 2022-08-31
style(stepper): fix error status styles

JoeyDjr:
This PR fixes the error status style on the Stepper.

Before : 
image

After : 
image


src/ui/molecules/stepper/stepper.scss
```diff
@@ -86,11 +86,11 @@ $colors: (
     overflow-y: auto;
     width: 100%;
     padding: 2px;
+  }
 
-    &.error {
-      @include with-theme() {
-        border: solid themed(error);
-      }
+  &.error {
```
lolottetheclash, line 91: I would have just moved the `error` classname above `.okp4-stepper-step-content` because it still refers to `.okp4-stepper-body`

src/ui/molecules/stepper/stepper.scss
```diff
@@ -81,17 +81,17 @@ $colors: (
     box-shadow: hsl(206 22% 7% / 35%) 0 10px 38px -10px, hsl(206 22% 7% / 20%) 0 10px 20px -15px;
   }
 
+  &.error {
+    @include with-theme() {
```
lolottetheclash, line 85: Also, I hadn't seen it before but I think you can remove `@include with-theme()` from the `error` because it is already present in `.okp4-stepper-body`. 
And since the `error` is from .`okp4-stepper-body` it can benefit from it.
We could move it just below `.okp4-stepper-body` by the way, because it is usually just below the classname.

lolottetheclash (approved):
Good for me !

src/ui/molecules/stepper/stepper.scss
```diff
@@ -66,39 +66,37 @@ $colors: (
 }
 
 .okp4-stepper-body {
```
fredericvilcot, line 68: Why `.okp4-stepper-body` is not inside `.okp4-stepper-main` as you have done for `.okp4-stepper-header`?
IMHO, the best way to use the `with-theme()` mixin is to wrap the whole selectors into the `.okp4-stepper-main` like this: 
```css
.okp4-stepper-main {
  @include with-theme() {
    ... /** all the others rules and selectors ho here **/
  }
}
```
If we have an issue with `stylelint` rules, we can see if it's relevant to increase the selector's depth.

src/ui/molecules/stepper/Stepper.tsx
```diff
@@ -269,12 +269,12 @@ export const Stepper: React.FC<StepperProps> = ({
...
-          className={classNames('okp4-stepper-step-content', {
-            error: state.stepsStatuses.get(state.activeStepIndex) === 'invalid'
-          })}
-        >
+      <div
+        className={classNames('okp4-stepper-body', {
+          error: state.stepsStatuses.get(state.activeStepIndex) === 'invalid'
+        })}
+      >
+        <div className={classNames('okp4-stepper-step-content')}>
```
ErikssonJoakim, line 277: No need for classNames() here.

```suggestion
        <div className="okp4-stepper-step-content">
```

ErikssonJoakim (approved):
Looking good just found an eye soar that could be nice to remove. Either way I approve 😄 👍

src/ui/molecules/stepper/stepper.scss
JoeyDjr: Yes that's right, from what I remember, it was a stylelint problem.

fredericvilcot: Please, try it and tell me back

JoeyDjr: It is indeed this error.

```
src/ui/molecules/stepper/stepper.scss
 40:15  ✖  Expected nesting depth to be no more than 6                                                       max-nesting-depth              
 40:15  ✖  Expected ".okp4-stepper-main .okp4-stepper-header .okp4-stepper-progress                          selector-max-compound-selectors
           .okp4-stepper-step-label.#{$color-key} div" to have no more than 4 compound selectors        
```

JoeyDjr: I have increased the max compound selectors and max nesting depth

fredericvilcot (approved):
Looks good!

PR #500, opened 2022-09-02
docs(text field): fix value control

JoeyDjr:
This PR fix an issue in the `TextField` story.
Before the change, it was not possible to control the `value` of the `TextField` and the change of the `defaultValue` was not visible.

_Before :_ 
image

_After :_ 
image

### DIFF
--- a/stories/atoms/components/textField/TextField.stories.mdx
+++ b/stories/atoms/components/textField/TextField.stories.mdx
@@ -54,10 +54,6 @@ import '../component.scss'
     },
     rightIcon: {
       control: false
-    },
-    value: {
-      control: { type: 'select' },
-      options: ['string', ['array', 'of', 'string']]
     }
   }}
   args={{


stories/atoms/components/textField/TextField.stories.mdx
```diff
@@ -54,10 +54,6 @@ import '../component.scss'
     },
     rightIcon: {
       control: false
-    },
-    value: {
-      control: { type: 'select' },
-      options: ['string', ['array', 'of', 'string']]
```
fredericvilcot, line 60: I have some interrogations about the purpose of this change:

- the `defaultValue` can't change after being set, that's a correct behavior: the default value can only define the `input` value at the component mount cycle.
- by doing this, you have removed the control that explained the user that the `value` could be either a `string` or a `string[]`. Using a control type as `select` can do the trick to simulate both types. Now, after your change, the control is an `object` that is unusable for the user.

I suggest to keep the control as `{ type: 'select' }`, or if you want to improve it we must be able to give the user the capability to use both types.

JoeyDjr: Ok, so I think we can improve the control of this field later. In the current use, it seems strange to me that you can't freely enter a value.

fredericvilcot: Yes, but how do you handle the `object` control here?

JoeyDjr: With the modification, you can simply type `"Foo"` or `["foo", "bar"]` in `Set object` to enter the value you want (string or list)

fredericvilcot: Ok but, I think it's not correct to expose the type control as `object`, even if everything is `object` in JS, we need to improve that :)

JoeyDjr: Yes I agree

fredericvilcot:
@JoeyDjr 
What about this one?
If not needed for now, we should consider closing the PR.

JoeyDjr:
No need for now.

PR #510, opened 2022-09-07
Fix/graphql ws errors

fredericvilcot:
This PR fixes the way errors are handled when `graphql-ws` client cannot open the socket.

It consists in adding event listeners that can return either a `Client` or an `Error` in an asynchronous way.

For that, the `HTTPFaucetGateway` has been updated to:

- handle errors thrown by the client 
- handle asynchronous subscription
- ensure that the socket is systematically closed when server has responded

⚠️ The test suite associated to `HTTPFaucetGateway` has been removed due to the complexity to mock the whole `graphql-ws`. The best way to test the adapter will be implementing an E2E test with a complete pipeline.


amimart (approved):
🤮

ccamel (approved):
image

PR #513, opened 2022-09-08
Feat/add size x-small to logo

ErikssonJoakim:
This PR adds the `size` **x-small** to the Logo component **and makes small changes to the sizes and Types descriptions.**

the **x-small** `size` follows the scale down from other sizes.

Current sizes story: 
Capture d’écran 2022-09-08 à 12 19 24

PR sizes story:
Capture d’écran 2022-09-08 à 12 18 53

Current types story description:
Capture d’écran 2022-09-08 à 12 23 46

PR types story description:
Capture d’écran 2022-09-08 à 12 23 27

### DIFF
--- a/src/ui/atoms/logo/Logo.tsx
+++ b/src/ui/atoms/logo/Logo.tsx
@@ -10,7 +10,7 @@ export type LogoProps = Readonly<{
   /**
    * The size of the logo, must be determined according to its use and location.
    */
-  readonly size?: 'small' | 'medium' | 'large'
+  readonly size?: 'x-small' | 'small' | 'medium' | 'large'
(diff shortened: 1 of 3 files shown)


lolottetheclash (commented):
I have a remark concerning the type `slogan`: I find that the size XS is unreadable (too small)
I think you should start with larger sizes for this type. What do you think?

lolottetheclash (approved):
Many thanks :)

ErikssonJoakim:
@lolottetheclash I followed the scale down of previous sizes but you're right it was too small. increased the size for slogan hope you agree with the new size 😄

Issue #494, opened 2022-09-01
List and ListItem improvements


ErikssonJoakim:
### **Purpose**

We'd like to improve **List** and **ListItem** by adding choice to the layout of **List** and by adding a `onClick` property to **ListItem**.

As a first implementation, **List** could have a layout `grid` and a layout `line`, line being the current layout.

We also want to change the property names `leftElement` and `rightElement` to correctly reflect their position in the component.

### **Visual**

![Capture d’écran 2022-09-01 à 13 53 05](https://user-images.githubusercontent.com/75730728/187907635-1f6aa837-5d56-495f-a7bf-6bb5da2206ad.png)

### **Use cases**

ListItems should be able to be shown in a grid layout or in a list.

It should be possible to attach a on click event to a **ListItem**.

#### **Properties**

**List**
- layout?: `grid` | `list` - default should be `list`

**ListItem**
- onClick?: `() => void`
- firstElement?: `Node`
- lastElement?: `Node`

### **Notes**

For a future feature we should also consider adding a `categoryColors` property for **List** and a `category` for **ItemList**

PR #514, opened 2022-09-08
Feat/improve list and listitem

ErikssonJoakim:
This PR implements #494 

There are some refactoring of documentation and components but it has not been a priority in this PR.
A lot of the changes concern the style of the components so don't hesitate to play around with it in your local server.
This marks a breaking change for `ListItem`.

**Layout story in `ListItem`**

Capture d’écran 2022-09-30 à 02 01 26

**On click story with hover on the clickable `ListItem`**
Capture d’écran 2022-09-30 à 02 01 48

Capture d’écran 2022-09-30 à 02 01 55

### DIFF
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -16,11 +16,15 @@ export type ListItemProps = Pick<ListProps, 'layout'> & {
    */
   readonly description?: string | JSX.Element
   /**
-   * An element that can provide visual information or allow interactions on the left side.
+   * An element that can provide visual information or allow interactions.
+   * If layout is in list mode, it will be displayed on the left side.
+   * If layout is in grid mode, it will be displayed on the top right corner.
    */
   readonly firstElement?: JSX.Element
   /**
-   * An element that can provide visual information or allow interactions on the right side.
+   * An element that can provide visual information or allow interactions.
+   * If layout is in list mode, it will be displayed on the right side.
+   * If layout is in grid mode, it will be displayed at the bottom.
    */
   readonly lastElement?: JSX.Element
   /**
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -4,12 +4,13 @@ import { Typography } from '../typography/Typography'
 import './listItem.scss'
 import { isString } from 'utils'
 import classNames from 'classnames'
+import type { ListProps } from '../list/List'
 
-export type ListItemProps = {
+export type ListItemProps = Pick<ListProps, 'layout'> & {
   /**
    * The main title of the item.
    */
-  readonly title?: string
+  readonly title?: string | JSX.Element
   /**
    * Additional information about the item
    */
@@ -29,37 +30,60 @@ export type ListItemProps = {
   readonly onClick?: () => void
 }
 
+type TitleProps = Pick<ListItemProps, 'title'>
+type DescriptionProps = Pick<ListItemProps, 'description'>
+
+const Title: React.FC<TitleProps> = ({ title }: DeepReadonly<TitleProps>) => {
+  if (!title) return null
+
+  const titleTypography = (
+    <div className="okp4-listitem-title">
+      <Typography color="inverted-text" fontSize="small" fontWeight="bold">
+        {title}
+      </Typography>
+    </div>
+  )
+
+  return isString(title) ? titleTypography : <div className="okp4-listitem-title">{title}</div>
+}
+
+const Description: React.FC<DescriptionProps> = ({
+  description
+}: DeepReadonly<DescriptionProps>) => {
+  if (!description) return null
+
+  const descriptionTypography = (
+    <div className="okp4-listitem-description">
+      <Typography color="inverted-text" fontSize="small">
+        {description}
+      </Typography>
+    </div>
+  )
+
+  return isString(description) ? (
+    descriptionTypography
+  ) : (
+    <div className="okp4-listitem-description">{description}</div>
+  )
+}
 export const ListItem: React.FC<ListItemProps> = ({
   title,
   description,
   firstElement,
   lastElement,
-  onClick
+  onClick,
+  layout
 }: DeepReadonly<ListItemProps>): JSX.Element => {
-  const renderDescription = (): string | JSX.Element | undefined =>
-    description && (
-      <div className="okp4-listitem-description">
-        {isString(description) ? (
-          <Typography as="div" color="inverted-text" fontSize="small">
-            {description}
-          </Typography>
-        ) : (
-          description
-        )}
-      </div>
-    )
-
+  const className = classNames('okp4-listitem-main', {
+    'layout-list': layout === 'list',
+    'layout-grid': layout === 'grid',
+    clickable: !!onClick
+  })
   return (
-    <div className={classNames('okp4-listitem-main', { clickable: !!onClick })} onClick={onClick}>
+    <div className={className} onClick={onClick}>
       {firstElement && <div className="okp4-listitem-first-element">{firstElement}</div>}
-      {title && (
-        <div className="okp4-listitem-title">
-          <Typography as="div" color="inverted-text" fontSize="small" fontWeight="bold">
-            {title}
-          </Typography>
-        </div>
-      )}
-      {renderDescription()}
+      <Title title={title} />
+      <Description description={description} />
       {lastElement && <div className="okp4-listitem-last-element">{lastElement}</div>}
     </div>
   )
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -3,6 +3,7 @@ import type { DeepReadonly } from 'superTypes'
 import { Typography } from '../typography/Typography'
 import './listItem.scss'
 import { isString } from 'utils'
+import classNames from 'classnames'
 
 export type ListItemProps = {
   /**
@@ -21,13 +22,19 @@ export type ListItemProps = {
    * An element that can provide visual information or allow interactions on the right side.
    */
   readonly lastElement?: JSX.Element
+  /**
+   * The callback function called when the list item is clicked.
+   * The `onClick` function is an event handler attached to the list item.
+   */
+  readonly onClick?: () => void
 }
 
 export const ListItem: React.FC<ListItemProps> = ({
   title,
   description,
   firstElement,
-  lastElement
+  lastElement,
+  onClick
 }: DeepReadonly<ListItemProps>): JSX.Element => {
   const renderDescription = (): string | JSX.Element | undefined =>
     description && (
@@ -43,7 +50,7 @@ export const ListItem: React.FC<ListItemProps> = ({
     )
 
   return (
-    <div className="okp4-listitem-main">
+    <div className={classNames('okp4-listitem-main', { clickable: !!onClick })} onClick={onClick}>
       {firstElement && <div className="okp4-listitem-first-element">{firstElement}</div>}
       {title && (
         <div className="okp4-listitem-title">
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -16,18 +16,18 @@ export type ListItemProps = {
   /**
    * An element that can provide visual information or allow interactions on the left side.
    */
-  readonly leftElement?: JSX.Element
+  readonly firstElement?: JSX.Element
   /**
    * An element that can provide visual information or allow interactions on the right side.
    */
-  readonly rightElement?: JSX.Element
+  readonly lastElement?: JSX.Element
 }
 
 export const ListItem: React.FC<ListItemProps> = ({
   title,
   description,
-  leftElement,
-  rightElement
+  firstElement,
+  lastElement
 }: DeepReadonly<ListItemProps>): JSX.Element => {
   const renderDescription = (): string | JSX.Element | undefined =>
     description && (
@@ -44,7 +44,7 @@ export const ListItem: React.FC<ListItemProps> = ({
 
   return (
     <div className="okp4-listitem-main">
-      {leftElement && <div className="okp4-listitem-left-element">{leftElement}</div>}
+      {firstElement && <div className="okp4-listitem-first-element">{firstElement}</div>}
       {title && (
         <div className="okp4-listitem-title">
           <Typography as="div" color="inverted-text" fontSize="small" fontWeight="bold">
@@ -53,7 +53,7 @@ export const ListItem: React.FC<ListItemProps> = ({
         </div>
       )}
       {renderDescription()}
-      {rightElement && <div className="okp4-listitem-right-element">{rightElement}</div>}
+      {lastElement && <div className="okp4-listitem-last-element">{lastElement}</div>}
     </div>
   )
 }
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -70,7 +70,8 @@ export const ListItem = ({
   layout = 'list'
 }: DeepReadonly<ListItemProps>): JSX.Element => {
   const className = classNames(`okp4-listitem-main layout-${layout}`, {
-    clickable: !!onClick
+    clickable: !!onClick,
+    'with-description': !!description
   })
   return (
     <div className={className} onClick={onClick}>
--- a/src/ui/atoms/listItem/ListItem.tsx
+++ b/src/ui/atoms/listItem/ListItem.tsx
@@ -4,9 +4,8 @@ import { Typography } from '../typography/Typography'
 import './listItem.scss'
 import { isString } from 'utils'
 import classNames from 'classnames'
-import type { ListProps } from '../list/List'
 
-export type ListItemProps = Pick<ListProps, 'layout'> & {
+export type ListItemProps = {
   /**
    * The main title of the item.
    */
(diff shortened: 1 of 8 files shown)


lolottetheclash (changes requested):
This is looking great!
I have a few comments, we can discuss it together next week :)

some additional comments :
- in the List.stories I think you can remove 
`          <div
            style={{
              display: 'grid',
              gap: '25px'
            }}
          >`
- in the ListItem.stories you have to add a wrapping `<List>` otherwise we don't get the grid layout and we end up with this className: `'okp4-listitem-main layout-undefined`

lolottetheclash (changes requested):

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,13 +135,200 @@ The `description` property allows you to provide additional information. It can
   </Story>
 </Canvas>
 
-## Right and left elements
+## Grid layout
+
+If used within a List component the style of the listItem can be changed to a grid style.
```
lolottetheclash, line 140: I'm not sure I understand what you mean in this description...

lolottetheclash (changes requested):

stories/atoms/components/listItem/ListItem.stories.mdx
ErikssonJoakim: If the component (`ListItem`) is a child of `List` then the style of the `ListItem` risks being changed depending on the `List` layout. It's important to be aware since otherwise you might forget to add a style for the grid layout.

src/ui/atoms/listItem/listItem.scss
```diff
@@ -1,40 +1,93 @@
...
+
+      .okp4-listitem-last-element {
+        grid-area: lastElement;
+        padding: 0 20px 20px;
+        align-self: end;
+        height: fit-content;
+      }
+    }
+
+    @keyframes hover-animation {
```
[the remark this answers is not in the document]
ErikssonJoakim: Thank you :)

src/ui/atoms/listItem/listItem.scss
```diff
@@ -1,40 +1,93 @@
...
+    overflow: auto;
 
-  .okp4-listitem-title {
-    grid-area: title;
-    justify-content: start;
-    margin: 3px;
-  }
+    &.clickable:hover {
+      cursor: pointer;
+      box-shadow: 0 2px 4px themed(primary-brand);
```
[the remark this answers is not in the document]
ErikssonJoakim: For me the `box-shadow` is visible in light mode. What browser are you using? Or do you mean in dark mode?
In dark mode the `box-shadow` is dark blue so with the current story` background-color` it's not so visible. 

I had trouble deciding what colour would be appropriate so if you have any suggestion of colour I'm happy to hear it 😃

stories/atoms/components/listItem/ListItem.stories.mdx
fredericvilcot: @ErikssonJoakim I think you should reword a bit this sentence, because, if @lolottetheclash had some difficulties to clearly understand it, let's imagine a lambda user / developer ...

src/ui/atoms/listItem/listItem.scss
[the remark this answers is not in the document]
lolottetheclash: yes it was on the dark theme! you have to find the right colors combination ;)

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -34,12 +34,14 @@ export type ListItemProps = Pick<ListProps, 'layout'> & {
...
-type TitleProps = Pick<ListItemProps, 'title'>
-type DescriptionProps = Pick<ListItemProps, 'description'>
+type TitleProps = {
+  title: string | JSX.Element
+}
+type DescriptionProps = {
+  description: string | JSX.Element
+}
 
 const Title: React.FC<TitleProps> = ({ title }: DeepReadonly<TitleProps>) => {
```
lolottetheclash, line 44: Can you please change `React.FC` type with `JSX.Element`?

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -34,12 +34,14 @@ export type ListItemProps = Pick<ListProps, 'layout'> & {
   readonly onClick?: () => void
 }
 
-type TitleProps = Pick<ListItemProps, 'title'>
-type DescriptionProps = Pick<ListItemProps, 'description'>
+type TitleProps = {
```
lolottetheclash, line 37: I think we should create a generic type for `string | JSX.Element` because we use it very often. 
We could put it in `superTypes.ts`, what do you think @fredericvilcot ?

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -78,16 +78,17 @@ export const ListItem: React.FC<ListItemProps> = ({
...
-    'layout-grid': layout === 'grid',
+  const className = classNames(`okp4-listitem-main layout-${layout}`, {
     clickable: !!onClick
   })
   return (
     <div className={className} onClick={onClick}>
       {firstElement && <div className="okp4-listitem-first-element">{firstElement}</div>}
-      <Title title={title} />
-      <Description description={description} />
+
```
lolottetheclash, line 87: Can you please remove the empty lines here?

src/ui/atoms/listItem/listItem.scss
```diff
@@ -1,40 +1,93 @@
...
-    'leftElement description rightElement';
-
   @include with-theme {
     background-color: themed(inverted-primary-background);
+    display: grid;
     border-radius: 10px;
-  }
+    column-gap: 10px;
+    padding: 10px;
+    overflow: auto;
```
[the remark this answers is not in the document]
lolottetheclash: Update: This bug happens only when we connect an external screen ...😳

lolottetheclash (changes requested):
Nice work @ErikssonJoakim  & @RaphaelEscrig !

@RaphaelEscrig, you should change the behavior of this piece of code: https://github.com/okp4/ui/blob/93dcdefd36f1b4497319a00fa4708e240090efd3/src/ui/atoms/listItem/ListItem.tsx#L50
Indeed, if the user submits a string, it is perfectly acceptable that it is rendered in a typography element. On the other hand, if he submits a JSX element, an image for example, it does not make sense that it is in a Typography element.
You can use a utility function `isString` in utils to render one or the other :)

stories/atoms/components/list/List.stories.mdx
```diff
@@ -90,3 +106,162 @@ These expected elements are [ListItem](/docs/atoms-listitem--overview) component
 ## Properties
 
 <ArgsTable story="Overview" />
+
+## Layout
+
+To allow for choice in the list `layout` you can choose between the following:
```
lolottetheclash, line 112: Can you add a "," after `layout` please?

stories/atoms/components/listItem/ListItem.stories.mdx
lolottetheclash, line 140: I think the overall explanation here needs to be modified to make it more fluid and clear. And be careful with the typography :
 `that it the`=> `that if the`
`the selectionned layout` => `the selected`

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -3,57 +3,93 @@ import type { DeepReadonly } from 'superTypes'
...
+  readonly title?: string | JSX.Element
   /**
    * Additional information about the item
    */
   readonly description?: string | JSX.Element
   /**
-   * An element that can provide visual information or allow interactions on the left side.
+   * An element that can provide visual information or allow interactions.
+   * If layout is in list mode, it will be displayed on the left side.
+   * If layout is in grid mode, it will be displayed on the top right corner.
```
lolottetheclash, line 21: Maybe the description here should be shortened a bit because the firstElement (or lastElement) is not complex to understand, and it is a lot of text rendered in the Storybook properties table.

lolottetheclash (changes requested):

src/ui/atoms/listItem/ListItem.tsx
ccamel: Ok for a generic type if we can find a good name for it (and you know, naming things is hard). But don't put it in the superTypes file because it should not be tainted with UI stuffs (look at the position of the file in the hierarchy)

lolottetheclash: Yes you're right! Each time I have the impression that it is under `UI`folder!
We can consider the creation of a `types.ts` file under `UI`.

src/ui/atoms/listItem/listItem.scss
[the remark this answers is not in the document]
ErikssonJoakim: @ccamel @fredericvilcot @JoeyDjr Any idea as to why this is ?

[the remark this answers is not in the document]
RaphaelEscrig: I don't have this bug with my two externals screens. Tested on Brave, Safari, Chrome and Firefox. (on MacOs Monterey) 😕

src/ui/atoms/listItem/ListItem.tsx
ErikssonJoakim: I agree maybe you could just precise that its positioning is dependent on the `layout` property value ?

src/ui/atoms/listItem/ListItem.tsx
ErikssonJoakim: Sounds good to me mind making an issue for this (out of scope of this PR)? @lolottetheclash 

@RaphaelEscrig I don't really understand why you are separating the types? For me the types are dependent on the types of ListItemProps

lolottetheclash: @ErikssonJoakim for me we can create a type in a file and apply it to `ListItem`, the idea is not to apply the changes to all other files. The refactor will be done in another specific PR.

RaphaelEscrig: @ErikssonJoakim, I did it because with the `TitleProps = Pick<ListItemProps, 'title'>`, title could be `string | JSX.Element | undefined`. It wasn't correct because we render this component only if title isn't undefined (new conditional rendering on the `<ListItem />` component) so the props received on the <Title /> could never be undefined.

fredericvilcot: TypeScript provides a high level typing to reverse optional to required:

```typescript
TitleProps = Required <Pick<ListItemProps, 'title'>>
```

lolottetheclash: simple and elegant solution!

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -3,57 +3,82 @@ import type { DeepReadonly } from 'superTypes'
...
 
+  return <div className="okp4-listitem-description">{description}</div>
+}
+export const ListItem = ({
+  title,
+  description,
+  firstElement,
+  lastElement,
+  onClick,
+  layout
```
ErikssonJoakim, line 71: To avoid having `undefined` as value we should pass the default value
```suggestion
  layout = 'list'
```

stories/atoms/components/list/List.stories.mdx
```diff
@@ -90,3 +106,162 @@ These expected elements are [ListItem](/docs/atoms-listitem--overview) component
...
 
 <ArgsTable story="Overview" />
+
+## Layout
+
+To allow for choice in the list `layout`, you can choose between the following:
+
+- **list** (default): This is a classic list with items being stacked upon each other tightly.
+  It gives the user an easy and fast way of comparing items.
+- **grid**: organizes the items in 350px \* 210px grids. The grid `layout` allow for more detailed list items.
```
fredericvilcot, line 116:
```
The grid `layout` allow for more detailed list items.
```

I can't understand clearly this sentence ...

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,13 +135,200 @@ The `description` property allows you to provide additional information. It can
...
+        }
+      />
+    </List>
+  </Story>
+</Canvas>
+
+## On click
+
+The `onClick` property makes the listItem clickable with a subtle animation and a change in cursor on hover.
+This facilitate its use by users that are used to clear visual cues as to when something is clickable.
```
fredericvilcot, line 242:
```suggestion
This facilitates its use by users that are used to clear visual cues as to when something is clickable.
```

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,13 +135,200 @@ The `description` property allows you to provide additional information. It can
...
+
+If a listItem does not have a `onClick` property it will not have this behaviour.
+
+<Canvas>
+  <Story name="On click">
+    {() => {
+      const handleClick = () => {
+        window.open('https://okp4.network/')
```
fredericvilcot, line 250: I would avoid this behavior as the user may not want to leave the current storybook to another page... (frustration)

stories/atoms/components/list/List.stories.mdx
```diff
@@ -90,3 +106,162 @@ These expected elements are [ListItem](/docs/atoms-listitem--overview) component
...
+
+<Canvas>
+  <Story name="Layout">
```
ErikssonJoakim, line 119: Would be good to refactor this code to make it more readable. You could just make a `ListItems` object containing the property values and then do a map for the two different layouts

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,13 +135,200 @@ The `description` property allows you to provide additional information. It can
...
-The `leftElement` and `rightElement` properties allow you to provide any `JSX.Element` as icons, buttons, text, etc. to an item in the list.  
+The `firstElement` and `lastElement` properties allow you to provide any `JSX.Element` as icons, buttons, text, etc. to an item in the list.  
 Theses properties can be used to provide visual information or add interactions with an item in the list.
 
 <Canvas>
-  <Story name="Right and left elements">
+  <Story name="First and last elements">
     {() => {
       const [open1, setOpen1] = useState(false)
       const handleOpenChange1 = () => setOpen1(false)
```
fredericvilcot, line 334: What do all these states and handlers mean? 🤔

fredericvilcot (changes requested):
That's a great job, especially about the main layout that was not easy to handle!
I had some remarks, please have a look on it to make the code more clearer 🙏

stories/atoms/components/listItem/ListItem.stories.mdx
ErikssonJoakim: I agree technically even without `List` the style can be changed by passing a `layout` value in the `ListItem`

ErikssonJoakim (commented):
Looking good! Just got a few remarks that needs looking into ;)

fredericvilcot:
@RaphaelEscrig @ErikssonJoakim 

Please, coordinate yourselves to implement the required changes and then we should merge it, because it's not good to keep opened PR for a long time without any action... 
Thanks 🙏

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -62,27 +62,39 @@ The `ListItem` allows to display an element of a [List](/docs/atoms-list--overvi
...
       control: false,
       table: {
         type: { summary: 'JSX.Element' }
       }
+    },
+    layout: {
+      description: 'The layout props from the List component.'
+    },
+    onClick: {
+      control: 'boolean'
```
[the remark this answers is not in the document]
ErikssonJoakim: Yes maybe it's not worth having as it could be confusing to the user?

I added the control to be able to show the hover behaviour but as `onClick` is not a boolean it might be better to just omit the control and have the user explore the `onClick` story to see this.

What do you guys think @lolottetheclash @JoeyDjr ?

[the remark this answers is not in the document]
fredericvilcot: Yes, please, remove the control as `onClick` is just a method.

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,20 +134,211 @@ The `description` property allows you to provide additional information. It can
   </Story>
 </Canvas>
 
-## Right and left elements
+## Grid layout
+
+The style of the listItem can be changed by passing either **list** (default) or **grid** as `layout` value.
```
lolottetheclash, line 139: I think this story is not clear enough...
Indeed, the `list` or `grid` value must be passed to the parent `List` element in order to be retrieved by the child. 
Here one might think that it is passed directly to `ListItem`.
Moreover, in the description it says "The style of the listItem can be changed by passing either list (default) or grid", but we only see `ListItem` arranged in grid layout.
Only the content of the `ListItem` changes, which does not fit with the description.

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -3,57 +3,79 @@ import type { DeepReadonly } from 'superTypes'
...
-   * An element that can provide visual information or allow interactions on the left side.
+   * An element that can provide visual information or allow interactions.
+   * The position of this element depends on the layout property value.
    */
-  readonly leftElement?: JSX.Element
+  readonly firstElement?: JSX.Element
   /**
-   * An element that can provide visual information or allow interactions on the right side.
+   * An element that can provide visual information or allow interactions.
+   * The position of this element depends on the layout property value.
```
fredericvilcot, line 25: What's the purpose of repeating this same sentence two times (for `firstElement` and `lastElement`)?

Please, add some specific context about their spatialization because, when I read that, the two props do exactly the same thing.

lolottetheclash (commented):
Nice refactor 👍
Just a few remarks though 😉
On the `ListItem` story: 
=> In overview, when you change `layout` into `grid`, the `ListItem` is not correctly displayed: the item is as long as a list item but the inside is arranged differently. 
It doesn't look like a 350px * 210px card .
 
=> title part: you should add ` It can be either a string or a JSX.Element.`as the description part.

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -123,20 +134,211 @@ The `description` property allows you to provide additional information. It can
...
+        </>
+      )
+    }}
+  </Story>
+</Canvas>
+
+## First and last elements
 
-The `leftElement` and `rightElement` properties allow you to provide any `JSX.Element` as icons, buttons, text, etc. to an item in the list.  
+The `firstElement` and `lastElement` properties allow you to provide any `JSX.Element` as icons, buttons, text, etc. to an item in the list.
```
fredericvilcot, line 332: It bothers me that you never speak about the position of each element, so, except their name, what's the real difference between those 2 properties?

fredericvilcot (changes requested):
Thanks for the changes 😌
I wrote down some remarks, I let you make the changes and then we need to merge it quickly please 🙏

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -62,27 +62,38 @@ The `ListItem` allows to display an element of a [List](/docs/atoms-list--overvi
         type: { summary: 'string | JSX.Element' }
```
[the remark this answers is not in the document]
ErikssonJoakim: Good catch! 👍 I will do this and in the upcoming refactor we can change the `JSX.Element` to `Node`

stories/atoms/components/listItem/ListItem.stories.mdx
ErikssonJoakim: This is because it's still possible to change the layout style of an item by passing the property directly. I think it's good to keep the description as is in order to separate `ListItem` from `List`.

For the story I agree I will implement a list layout as well.

stories/atoms/components/listItem/ListItem.stories.mdx
ErikssonJoakim: I'll add some description about their position 😺

src/ui/atoms/listItem/ListItem.tsx
```diff
@@ -3,57 +3,79 @@ import type { DeepReadonly } from 'superTypes'
...
    */
-  readonly rightElement?: JSX.Element
+  readonly lastElement?: JSX.Element
+  /**
+   * The callback function called when the list item is clicked.
+   * The `onClick` function is an event handler attached to the list item.
```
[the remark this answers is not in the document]
ErikssonJoakim: I took inspiration from the button component, but you're right it might be a bit over the top

lolottetheclash (changes requested):
⚠️ You have 2 broken links on "read more" list items.
And I think there is too much text content inside the list items of layout, it gives an impression of overload...

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -351,62 +371,88 @@ Theses properties can be used to provide visual information or add interactions
...
+                    onClick={() => setOpen2(true)}
+                  />
+                </div>
               }
             />
           </List>
           <Toast
             isOpened={open1}
-            onOpenChange={handleOpenChange1}
+            onOpenChange={() => setOpen1(false)}
```
lolottetheclash, line 449: Could you remove lambdas from global jsx/mdx and turn them into functions?

src/ui/atoms/listItem/listItem.scss
```diff
@@ -1,40 +1,84 @@
...
+    }
+
+    .okp4-listitem-last-element {
+      grid-area: lastElement;
+    }
 
-  .okp4-listitem-left-element {
-    grid-area: leftElement;
-    margin: 5px auto 5px 5px;
+    .okp4-listitem-first-element > img,
```
lolottetheclash, line 36: I'm not sure why you are targeting an `img` because I don't see any in the jsx 🤔

lolottetheclash (changes requested):
That's good, you're almost there!
A little remark: I don't think that the "simplify layout" commit is a breaking change. 
Before, the basic display was in list and no `layout` property was provided.
Now there is one, but it is not mandatory, it defaults to `list`. 
So, since it's optional, and the classic rendering is the same as before, I would say it's not a BC.

stories/atoms/components/listItem/ListItem.stories.mdx
ErikssonJoakim: I was not aware this is unwanted. It is now fixed

ErikssonJoakim:
@lolottetheclash After talking with @fredericvilcot I got some clarity to what is considered a breaking change and as you say a small change in style is not it. Thank you for your feedback.

src/ui/atoms/listItem/listItem.scss
```diff
@@ -1,40 +1,79 @@
...
 
-  .okp4-listitem-title {
-    grid-area: title;
-    justify-content: start;
-    margin: 3px;
-  }
+    &.clickable:hover {
+      cursor: pointer;
+      box-shadow: 0 2px 4px themed(inverted-field);
+      animation: 0.2s ease-out hover-animation;
```
RaphaelEscrig, line 16: Here you can do `animation: 0.2s ease-out hover-animation forwards;` . With the "forwards" at the end of the ligne, the state will persist once the animation is done. If you implement it, you can delete the line just below.

https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode

stories/atoms/components/listItem/ListItem.stories.mdx
```diff
@@ -92,14 +112,28 @@ The `ListItem` allows to display an element of a [List](/docs/atoms-list--overvi
...
 
 <Canvas>
-  <Story name="Title">{() => <ListItem title="The title of the item." />}</Story>
+  <Story name="Title">
+    {() => (
+      <List>
+        <ListItem title="The title as a string" />
+        <ListItem
+          title={
+            <Typography color="inverted-text" fontWeight="bold">
```
RaphaelEscrig, line 124: Here because just above you said "The title as a string", I would have written "The title as a node element" in order to fit more with the description.

fredericvilcot (approved):
LGTM, great!

src/ui/atoms/listItem/listItem.scss
ErikssonJoakim: Cool I didn't know about this property, thanks! 😃

lolottetheclash (changes requested):
Could you remove the 4 last lambdas in `ListItem.stories.mdx` please?
And it will be good for me ;)

lolottetheclash (approved):

RaphaelEscrig (approved):
🚀

JoeyDjr (approved):

PR #518, opened 2022-09-09
Feat/task progress

fredericvilcot:
👉 This PR adds a `progress` property to the existing `task` entity in order to handle all generic asynchronous operations.

🧠 This feature is built upon Event-Sourcing principles that allow to create / update the state with invariant events. These events bring the user the capability to configure the progress bounds (`min` / `max`) and to set the progress `current` value.

⚠️ Commands could be invoked with a corrupted payload, so the use-cases guarantee that the `progress` property in `task` entity will never be in an inconsistent state, and that by systematically check the values that are passed as parameters.

🧪 All use-cases are fully unit tested, as well as the `task` builder and the `progressInvariant` method used to check the values given as parameters.


src/domain/task/builder/task.builder.ts
```diff
@@ -78,7 +87,17 @@ export class TaskBuilder {
...
+      (this.task.progress ? this.progressInvariant(this.task.progress) : true)
+    )
+  }
+
+  private progressInvariant(progress: DeepReadonly<TaskProgress>): boolean {
+    return (
+      progress.current <= progress.max &&
+      progress.current >= progress.min &&
+      progress.max > progress.min &&
+      progress.min < progress.max
```
JoeyDjr, line 100:
```suggestion
      progress.max > progress.min
```
It is the same condition. Just one is enough

src/domain/task/usecase/register-task/registerTask.ts
```diff
@@ -5,17 +5,30 @@ import { ErrorMapper } from 'domain/error/mapper/error.mapper'
...
 const dispatchError = (error: unknown, dispatch: ReduxStore['dispatch']): void => {
   const errorToDispatch = ErrorMapper.mapRawErrorToEntity(error)
   dispatch(ThrowErrorActions.errorThrown(errorToDispatch))
 }
 
+const isProgressValid = (progress: DeepReadonly<Progress>): boolean =>
+  progress.current <= progress.max &&
+  progress.current >= progress.min &&
+  progress.max > progress.min &&
+  progress.min < progress.max
```
JoeyDjr, line 19: same here

src/domain/task/usecase/amend-task-progress/amendTaskProgress.ts
```diff
@@ -0,0 +1,45 @@
...
+const dispatchError = (error: unknown, dispatch: ReduxStore['dispatch']): void => {
+  const errorToDispatch = ErrorMapper.mapRawErrorToEntity(error)
+  dispatch(ThrowErrorActions.errorThrown(errorToDispatch))
+}
+
+const isProgressValid = (progress: DeepReadonly<Progress>): boolean =>
+  progress.current <= progress.max &&
+  progress.current >= progress.min &&
+  progress.max > progress.min &&
+  progress.min < progress.max
```
JoeyDjr, line 19:
```suggestion
  progress.max > progress.min
```
same here.

JoeyDjr (changes requested):
Great. Juste a small remark

src/domain/task/builder/task.builder.ts
fredericvilcot: of course 🤪

JoeyDjr (dismissed):

JoeyDjr (commented):
Look good to me. Just one remark.

src/domain/task/builder/task.builder.ts
```diff
@@ -78,7 +88,8 @@ export class TaskBuilder {
       this.task.lastUpdateDate instanceof Date &&
       this.task.type.length > 0 &&
       this.task.status.length > 0 &&
-      this.task.initiator.length > 0
+      this.task.initiator.length > 0 &&
+      (this.task.progress ? progressInvariant(this.task.progress) : true)
```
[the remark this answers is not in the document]
fredericvilcot: Yep, that's a good catch, thx :)

JoeyDjr (approved):

src/domain/task/usecase/configure-task-progress-bounds/configureTaskProgressBounds.ts
```diff
@@ -0,0 +1,61 @@
...
+  async (dispatch, getState) => {
+    const {
+      id,
+      timestamp = new Date(),
+      progressBounds
+    }: ConfigureTaskProgressBounds = configureTaskProgressBoundsPayload
+    if (!getState().task.byId.has(id)) {
+      dispatchError(
+        new UnspecifiedError(
+          `Oops.. The provided id '${id}' does not exist, so we can't configure the progress bounds of an unknown task..`
```
ccamel, line 30: Typo.
```suggestion
          `Oops.. The provided id '${id}' does not exist, so we can't configure the progress bounds of an unknown task...`
```

src/domain/task/usecase/configure-task-progress-bounds/configureTaskProgressBounds.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (Object.keys(progressBounds).length > 0) {
+      const progressEntity = getState().task.byId.get(id)?.progress
+      const progressFuture = { ...progressEntity, ...progressBounds }
+
+      if (!progressInvariant(progressFuture)) {
+        dispatchError(
+          new UnspecifiedError(
+            `Oops.. The provided progress bounds '${JSON.stringify(
+              progressBounds
+            )}' are not valid, so we can't configure the progress bounds..`
```
ccamel, line 46: Typo.
```suggestion
            )}' are not valid, so we can't configure the progress bounds...`
```

src/domain/task/utils/task.utils.ts
```diff
@@ -0,0 +1,17 @@
+import type { DeepReadonly } from 'superTypes'
+
+type Progress = { min?: number; max?: number; current?: number }
+
+export const progressInvariant = (progress: DeepReadonly<Progress>): boolean => {
```
ccamel, line 5: I would have called it: `isProgressValid`.

src/domain/task/utils/task.utils.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { progressInvariant } from './task.utils'
+
+describe('Considering the progressInvariant() function', () => {
+  describe.each`
+    current      | min          | max          | expectedResult
+    ${undefined} | ${undefined} | ${undefined} | ${true}
+    ${10}        | ${undefined} | ${undefined} | ${true}
```
ccamel, line 7: What's the meaning of this case? You can specify a progress value without having defined the bounds?

src/domain/task/usecase/set-task-progress-value/setTaskProgressValue.ts
```diff
@@ -0,0 +1,54 @@
...
+  async (dispatch, getState) => {
+    const {
+      id,
+      timestamp = new Date(),
+      progressValue
+    }: SetTaskProgressValue = setTaskProgressValuePayload
+    if (!getState().task.byId.has(id)) {
+      dispatchError(
+        new UnspecifiedError(
+          `Oops.. The provided id '${id}' does not exist, so we can't set the progress value of an unknown task..`
```
ccamel, line 28: Typo.
```suggestion
          `Oops.. The provided id '${id}' does not exist, so we can't set the progress value of an unknown task...`
```

src/domain/task/usecase/set-task-progress-value/setTaskProgressValue.ts
```diff
@@ -0,0 +1,54 @@
...
+      )
+      return
+    }
+
+    const progressEntity = getState().task.byId.get(id)?.progress
+    const progressFuture = { ...progressEntity, current: progressValue }
+    if (!progressInvariant(progressFuture)) {
+      dispatchError(
+        new UnspecifiedError(
+          `Oops.. The provided progress value '${progressValue}' is not valid, so we can't set the progress value..`
```
ccamel, line 40: Typo.
```suggestion
          `Oops.. The provided progress value '${progressValue}' is not valid, so we can't set the progress value...`
```

src/domain/task/store/reducer/task.reducer.ts
```diff
@@ -68,10 +73,57 @@ const removeTask = (
...
 }
 
+const configureTaskProgressBounds = (
+  state: DeepReadonly<TaskState>,
+  payload: DeepReadonly<TaskProgressBoundsConfiguredPayload>
+): TaskState => {
+  const { id, timestamp, progressBounds }: TaskProgressBoundsConfiguredPayload = payload
+  const foundTaskById = state.byId.get(id)
+  const { max, min }: Progress = progressBounds
+  const bounds = {
```
ccamel, line 83: There is a subtlety here. By doing this, we can update the bounds with concrete values, but we can never remove them, while your invariant accepts it. So we have an unattainable state.

It's not easy to manage, and I think we have to think a little bit before changing the code, which can be addressed in a subsequent PR.

ccamel (commented):
That's a good job. A few small comments that I'll leave you to consider or not.

ErikssonJoakim (approved):
LGTM 👍 Some challenging concepts to get your head around but a great brain exercise 🧠 😄

src/domain/task/store/reducer/task.reducer.ts
fredericvilcot: Yes, it's a tricky interrogation.
I see your point, but I think we could address it later because that modifies the way we use commands and we design the store.

src/domain/task/utils/task.utils.ts
fredericvilcot: Yep, much better

src/domain/task/utils/task.utils.spec.ts
fredericvilcot: Yep, as all properties are optional, I think it's kind of weird to prevent from setting a current value without bounds.
Maybe I'm wrong, but I think we may meet some use cases where we have to set initial current value before bounds.
If it's OK for you, I will preserve this behavior 😌

src/domain/task/store/reducer/task.reducer.ts
ccamel: Of course!

ccamel (approved):

PR #532, opened 2022-09-12
FilePicker : Add clearAll story

JoeyDjr:
This PR add `clearAll` docs to storybook.

image

### DIFF
--- a/stories/organisms/components/filePicker/FilePicker.stories.mdx
+++ b/stories/organisms/components/filePicker/FilePicker.stories.mdx
@@ -97,6 +97,34 @@ It is a combination of the [FileInput](/docs/atoms-fileinput--overview) and the 
 
 <ArgsTable story="Overview" />
 
+## Clear all
+
+The `clearAll` property allows the user to add a clickable text to delete all selected files.  
+This text appears automatically if at least two files have been selected.  
+By default the `clearAll` value is `true`.
+
+<Canvas>
+  <Story name="Clear all">
+    <div
+      style={{
+        display: 'flex'
+      }}
+    >
+      <FilePicker
+        label="Drop your files here, or browse"
+        clearAll={true}
+        description={
+          <div style={{ display: 'flex', flexDirection: 'column', alignItems: 'center' }}>
+            <Typography as="div" fontFamily="brand" fontSize="small" fontWeight="xlight">
+              Display `Clear all` to remove all the selected files.
+            </Typography>
+          </div>
+        }
+      />
+    </div>
+  </Story>
+</Canvas>
+
 ## Usage
 
 To configure the FilePicker, the [@okp4/ui](https://www.npmjs.com/package/@okp4/ui) library provides all the necessary elements. You will just have to follow these two steps:


stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -95,6 +95,34 @@ It is a combination of the [FileInput](/docs/atoms-fileinput--overview) and the
 
 <ArgsTable story="Overview" />
 
+## Clear all
+
+The `clearAll` property allows the user to add a link to delete all selected files.  
+This link appears automatically if at leat two files have been selected.
```
ccamel, line 101:
```suggestion
This link appears automatically if at least two files have been selected.
```

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -97,6 +97,34 @@ It is a combination of the [FileInput](/docs/atoms-fileinput--overview) and the
 
 <ArgsTable story="Overview" />
 
+## Clear all
+
+The `clearAll` property allows the user to add a link to delete all selected files.
```
lolottetheclash, line 102: For me, the term "link" is to be avoided because it could be thought that it acts as a real html `link` and would take the user to another page to delete the selected files. But I'm not sure if my remark is relevant !

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -97,6 +97,34 @@ It is a combination of the [FileInput](/docs/atoms-fileinput--overview) and the
 
 <ArgsTable story="Overview" />
 
+## Clear all
+
+The `clearAll` property allows the user to add a clickable text to delete all selected files.  
+This text appears automatically if at least two files have been selected.  
+By default the `clearAll` value is `true`.
```
fredericvilcot, line 104: I didn't notice that in the `FilePicker` implementation, but IMHO it would be better to revert the behavior and initialize the `clearAll` property to `false`.
In the interests of seeking the richest possible User eXperience, it is preferable not to force a behavior but rather to bring about ways to modify it.

fredericvilcot:
@JoeyDjr 
By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.
`clearAll ` refers to the action itself
`showClearAll` refers to the display of the button

ccamel:
> @JoeyDjr By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.

@fredericvilcot I agree with this. We should consider the change to be clean on the intent carried by this property.

ErikssonJoakim:
> @JoeyDjr By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.
> 
> `clearAll` refers to the action itself `showClearAll` refers to the display of the button
> 
> What do you think @ccamel @lolottetheclash @ErikssonJoakim ?

I agree but IMO something like `hasClearAll` makes more sense

lolottetheclash:
> > @JoeyDjr By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.
> > `clearAll` refers to the action itself `showClearAll` refers to the display of the button
> > What do you think @ccamel @lolottetheclash @ErikssonJoakim ?
> 
> I agree but IMO something like `hasClearAll` makes more sense

Or maybe `isClearable`?

stories/organisms/components/filePicker/FilePicker.stories.mdx
JoeyDjr: > > > @JoeyDjr By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.
> > > `clearAll` refers to the action itself `showClearAll` refers to the display of the button
> > > What do you think @ccamel @lolottetheclash @ErikssonJoakim ?
> > 
> > 
> > I agree but IMO something like `hasClearAll` makes more sense
> 
> Or maybe `isClearable`?

I think we loose the "all" aspect with this name

I think `showClearAll` is the best, or `displayClearAll`

JoeyDjr:
> > > @JoeyDjr By the way, I missed a comment in my review, I think we should rename also the `clearAll` property into `showClearAll` because the purpose is completely different depending on the name used.
> > > `clearAll` refers to the action itself `showClearAll` refers to the display of the button
> > > What do you think @ccamel @lolottetheclash @ErikssonJoakim ?
> > 
> > 
> > I agree but IMO something like `hasClearAll` makes more sense
> 
> Or maybe `isClearable`?

I think we loose the "all" aspect with this name

I think `showClearAll` is the best, or `displayClearAll`

stories/organisms/components/filePicker/FilePicker.stories.mdx
```diff
@@ -97,6 +97,36 @@ It is a combination of the [FileInput](/docs/atoms-fileinput--overview) and the
 
 <ArgsTable story="Overview" />
 
+## Clear all
+
+Need to delete all uploaded files at once?  
+We thought it would be handy to offer you a turnkey way to do it!  
+For that, you can enable the `showClearAll` property (by giving it `true`) and that's it!
+The displayed text is themed and planned to fit harmoniously above the list.
+By default the `showClearAll` value is `false`.
```
ErikssonJoakim, line 106: Just small remark but values should be in bold as specified by the [wiki](https://github.com/okp4/ui/wiki/Create-A-Story).

I would also like to take the opportunity to suggest some changes and adding description about the `showClearAll` internationalisation.

```suggestion
For that, you only have to pass `showClearAll` to the filePicker and that's it!
The displayed text is themed and internationalized (english and french) and fits harmoniously above the list.
By default the `showClearAll` value is **false**.
```

JoeyDjr: Ok
Will be done in this PR : https://github.com/okp4/ui/pull/534

PR #533, opened 2022-09-13
TextField : fix resize when full width

JoeyDjr:
This PR fixes the resizing problem when the TextField is full width by allowing only the height of the TextField to be resized.

image

### DIFF
--- a/src/ui/atoms/inputBase/InputBase.tsx
+++ b/src/ui/atoms/inputBase/InputBase.tsx
@@ -56,9 +56,9 @@ export type InputBaseProps = {
    */
   readonly disableAreaResize?: boolean
   /**
-   * If true, the user can only resize the height of the text area.
+   * If true, the input will take 100% of its parent's size.
    */
-  readonly resizeHeightOnly?: boolean
+  readonly fullWidth?: boolean
   /**
    * Whether input has a themed border.
    */
@@ -87,7 +87,7 @@ export const InputBase = ({
   multiline = false,
   numberOfLines,
   disableAreaResize = false,
-  resizeHeightOnly = false,
+  fullWidth = false,
   withBorder = false
 }: InputBaseProps): JSX.Element => {
   const containerClass = classNames(`okp4-input-base-container`, {
@@ -101,7 +101,7 @@ export const InputBase = ({
   })
   const textareaClass = classNames(inputClass, 'okp4-input-base-textarea', {
     'disable-resize': multiline && disableAreaResize,
-    'resize-height': resizeHeightOnly
+    'resize-height-only': fullWidth && multiline
   })
   const props = {
     defaultValue,
--- a/src/ui/atoms/inputBase/InputBase.tsx
+++ b/src/ui/atoms/inputBase/InputBase.tsx
@@ -56,6 +56,10 @@ export type InputBaseProps = {
    */
   readonly disableAreaResize?: boolean
   /**
+   * If true, the user can only resize the height of the text area.
+   */
+  readonly resizeHeightOnly?: boolean
+  /**
    * Whether input has a themed border.
    */
   readonly withBorder?: boolean
@@ -83,6 +87,7 @@ export const InputBase = ({
   multiline = false,
   numberOfLines,
   disableAreaResize = false,
+  resizeHeightOnly = false,
   withBorder = false
 }: InputBaseProps): JSX.Element => {
   const containerClass = classNames(`okp4-input-base-container`, {
@@ -95,7 +100,8 @@ export const InputBase = ({
     error: hasError
   })
   const textareaClass = classNames(inputClass, 'okp4-input-base-textarea', {
-    'disable-resize': multiline && disableAreaResize
+    'disable-resize': multiline && disableAreaResize,
+    'resize-height': resizeHeightOnly
   })
   const props = {
     defaultValue,
--- a/src/ui/atoms/textField/TextField.tsx
+++ b/src/ui/atoms/textField/TextField.tsx
@@ -16,20 +16,14 @@ export type TextFieldProps = InputBaseProps & {
    * Displays a message to the user below the input area.
    */
   readonly helperText?: string
-  /**
-   * If true, the TextField will take 100% of its parent's size.
-   */
-  readonly fullWidth?: boolean
 }
 
 export const TextField: React.FC<TextFieldProps> = ({
   size = 'medium',
   helperText,
-  fullWidth = false,
-  multiline = false,
-  disableAreaResize = false,
   ...props
 }: TextFieldProps): JSX.Element => {
+  const { multiline, disableAreaResize, fullWidth, hasError }: InputBaseProps = props
   const containerClass = classNames(
     `okp4-text-field-main ${multiline && !disableAreaResize ? 'auto' : size}`,
     {
@@ -39,16 +33,11 @@ export const TextField: React.FC<TextFieldProps> = ({
 
   return (
     <div className={containerClass}>
-      <InputBase
-        {...props}
-        disableAreaResize={disableAreaResize}
-        multiline={multiline}
-        resizeHeightOnly={fullWidth}
-      />
+      <InputBase {...props} />
       {helperText && (
         <Typography
           as="div"
-          color={props.hasError ? 'error' : 'info'}
+          color={hasError ? 'error' : 'info'}
           fontSize="x-small"
           fontWeight="bold"
           noWrap
--- a/src/ui/atoms/textField/TextField.tsx
+++ b/src/ui/atoms/textField/TextField.tsx
@@ -26,18 +26,25 @@ export const TextField: React.FC<TextFieldProps> = ({
   size = 'medium',
   helperText,
   fullWidth = false,
+  multiline = false,
+  disableAreaResize = false,
   ...props
 }: TextFieldProps): JSX.Element => {
   const containerClass = classNames(
-    `okp4-text-field-main ${props.multiline && !props.disableAreaResize ? 'auto' : size}`,
+    `okp4-text-field-main ${multiline && !disableAreaResize ? 'auto' : size}`,
     {
-      'full-width': !props.multiline && fullWidth
+      'full-width': fullWidth
     }
   )
 
   return (
     <div className={containerClass}>
-      <InputBase {...props} />
+      <InputBase
+        {...props}
+        disableAreaResize={disableAreaResize}
+        multiline={multiline}
+        resizeHeightOnly={fullWidth}
+      />
       {helperText && (
         <Typography
           as="div"
--- a/src/ui/atoms/textField/TextField.tsx
+++ b/src/ui/atoms/textField/TextField.tsx
@@ -20,44 +20,24 @@ export type TextFieldProps = InputBaseProps & {
    * If true, the TextField will take 100% of its parent's size.
    */
   readonly fullWidth?: boolean
-  /**
-   * If true, the TextField allows multiline text as a text area.
-   */
-  readonly multiline?: boolean
-  /**
-   * If multiline, defines the number of lines of the text area.
-   */
-  readonly numberOfLines?: number
-  /**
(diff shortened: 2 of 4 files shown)


stories/atoms/components/textField/TextField.stories.mdx
```diff
@@ -267,12 +267,23 @@ The `withBorder` property allows you to add a themed border to the textField and
 ## Multiline
 
 The `multiline` property allows you to write text on several lines. The simple text field becomes a text area.  
-When `multiline` is `true` and [disableAreaResize](/docs/atoms-textfield--disable-area-resize) is `false` the [size](/docs/atoms-textfield--sizes) is automatically `auto`.
+When `multiline` is `true` and [disableAreaResize](/docs/atoms-textfield--disable-area-resize) is `false`
+the [size](/docs/atoms-textfield--sizes) is automatically `auto`.  
+When `fullWidth` is `true`and [disableAreaResize](/docs/atoms-textfield--disable-area-resize) is `false`
```
ErikssonJoakim, line 272: The description is not consistent here and as such can lead to confusion as to if `multiline` is necessary for the behavior. I suggest reworking the description so that we don't need to repeat the `multiline` **value**.

On a side note be aware that properties are supposed to be in bac tics and not hyperlink text.

ErikssonJoakim (changes requested):
Looking good just got a small problem with the description that I'd like we take a look at 😃

src/ui/atoms/inputBase/InputBase.tsx
```diff
@@ -55,6 +55,10 @@ export type InputBaseProps = {
    * If true and multiline, the user can't resize the text area.
    */
   readonly disableAreaResize?: boolean
+  /**
+   * If true, the user can only resize the height of the text area.
+   */
+  readonly resizeHeightOnly?: boolean
```
fredericvilcot, line 61: I'm a little embarrassed about using this property.

Indeed, if I understand correctly, the purpose of this property is to prevent the resize of the text area when it is in full-width.

What bothers me is exposing a property to the user which, in the end, is only used for internal modifications to the component.

I think it should just be a condition carried by the `InputBase` component and attached to the `fullWidth` and `multiline` properties.

Thus, it shouldn't be exposed as `props`.

src/ui/atoms/textField/TextField.tsx
```diff
@@ -20,43 +20,30 @@ export type TextFieldProps = InputBaseProps & {
...
-   * If true and multiline, the user can't resize the text area.
-   */
-  readonly disableAreaResize?: boolean
 }
 
 export const TextField: React.FC<TextFieldProps> = ({
   size = 'medium',
   helperText,
   fullWidth = false,
   multiline = false,
```
fredericvilcot, line 29: These props are already exposed by the `InputBase` component, so you only need to spread them with the others.

src/ui/atoms/textField/TextField.tsx
```diff
@@ -20,43 +20,30 @@ export type TextFieldProps = InputBaseProps & {
...
 
   return (
     <div className={containerClass}>
       <InputBase
+        {...props}
         disableAreaResize={disableAreaResize}
         multiline={multiline}
-        numberOfLines={numberOfLines}
-        {...props}
+        resizeHeightOnly={fullWidth}
```
fredericvilcot, line 46: What about if the user set `resizeHeightOnly` to `false` event if you are in `fullWidth` mode?

There is no effect as the value is based upon `fullWidth`'s one .. So it can lead to a misunderstood for the user which is not good...

stories/atoms/components/textField/TextField.stories.mdx
```diff
@@ -267,12 +267,23 @@ The `withBorder` property allows you to add a themed border to the textField and
...
+the user can only resize the height of the textField.
```
fredericvilcot, line 273: We need to rework this whole paragraph as I can't clearly understand what it means...

If `X` is `true` and `Y` is `false`, so `Z` is ... But when `W` is `true` and `Y` is `false` etc... 

Sounds quite hard to understand easily 😅

fredericvilcot (changes requested):
I think we need to rework the way the resize is handled in the component, don't hesitate if you need more informations 😌

fredericvilcot:
@JoeyDjr Thanks for the changes, could you please update the description accordingly?

src/ui/atoms/inputBase/InputBase.tsx
```diff
@@ -55,6 +55,10 @@ export type InputBaseProps = {
    * If true and multiline, the user can't resize the text area.
    */
   readonly disableAreaResize?: boolean
+  /**
+   * If true, the TextField will take 100% of its parent's size.
```
fredericvilcot, line 59: If the `fullwidth` property is now part of the `InputBase` component, I will not mention `TextField` anymore in the property description as it can be used in any component that entends base props. 

```suggestion
   * If true, the input will take 100% of its parent's size.
```

src/ui/atoms/inputBase/InputBase.tsx
```diff
@@ -95,7 +100,8 @@ export const InputBase = ({
     error: hasError
   })
   const textareaClass = classNames(inputClass, 'okp4-input-base-textarea', {
-    'disable-resize': multiline && disableAreaResize
+    'disable-resize': multiline && disableAreaResize,
+    'resize-height': fullWidth && multiline
```
fredericvilcot, line 104: I will add a little bit more context to the class name because, by default, the text area can be resized vertically and horizontally.

```suggestion
    'resize-height-only': fullWidth && multiline
```

src/ui/atoms/textField/TextField.tsx
```diff
@@ -20,48 +20,28 @@ export type TextFieldProps = InputBaseProps & {
    * If true, the TextField will take 100% of its parent's size.
    */
   readonly fullWidth?: boolean
```
fredericvilcot, line 22: If you don't need to overload this property on the `TextField` component (and I think you don't), this must be removed as `fullWidth` is already spread with `InputBase` props.

src/ui/atoms/textField/TextField.tsx
```diff
@@ -20,48 +20,28 @@ export type TextFieldProps = InputBaseProps & {
...
-  numberOfLines,
-  disableAreaResize = false,
   ...props
 }: TextFieldProps): JSX.Element => {
+  const { multiline, disableAreaResize, fullWidth, hasError }: InputBaseProps = props
   const containerClass = classNames(
     `okp4-text-field-main ${multiline && !disableAreaResize ? 'auto' : size}`,
     {
-      'full-width': !multiline && fullWidth
+      'full-width': fullWidth
```
fredericvilcot, line 34: I don't know if it's possible, but I will expect this className to be carried directly by the `InputBase` base style. 
Tell me if I'm wrong 😌

fredericvilcot (changes requested):
Thanks for the changes, it's really much better 💪
I still have some remarks about the orchestration between `InputBase` component and its  descendants.

src/ui/atoms/textField/TextField.tsx
JoeyDjr: I think we can't because the `fullWidth` props is used by a container that contains the `InputBase` 

```tsx
const containerClass = classNames(
    `okp4-text-field-main ${multiline && !disableAreaResize ? 'auto' : size}`,
    {
      'full-width': fullWidth
    }
  )

  return (
    <div className={containerClass}>
      <InputBase {...props} />
      {helperText && (
        <Typography
          as="div"
          color={hasError ? 'error' : 'info'}
          fontSize="x-small"
          fontWeight="bold"
          noWrap
        >
          {helperText}
        </Typography>
      )}
    </div>
  )
  ```

Comment (JoeyDjr) [reply src/ui/atoms/textField/TextField.tsx:34]: What do you think @lolottetheclash ?

Comment (lolottetheclash) [reply src/ui/atoms/textField/TextField.tsx:30]: At first sight I would say that this classname must be managed here in order to wrap the input and the helperText. Otherwise, it would be necessary to refactor the whole thing in order to lower the management of the classname by one level.

Comment (fredericvilcot) [reply src/ui/atoms/textField/TextField.tsx:30]: OK I see the point, but, IMHO, the way of dealing with the `fullWidth` property is not correct.
Doing like this forces all descendants to declare their own way of handling this property, so there Is no genericity at all.
I think this should be part of a little refactoring that we should address soon if you are OK 😌

Comment (JoeyDjr) [reply src/ui/atoms/textField/TextField.tsx:30]: ok for me

File: src/ui/atoms/textField/TextField.tsx [hunk 10]
```diff
@@ -16,52 +16,28 @@ export type TextFieldProps = InputBaseProps & {
...
-   */
-  readonly multiline?: boolean
-  /**
-   * If multiline, defines the number of lines of the text area.
-   */
-  readonly numberOfLines?: number
-  /**
-   * If true and multiline, the user can't resize the text area.
-   */
-  readonly disableAreaResize?: boolean
```
lolottetheclash, line 34: On the other hand the fact of having removed the props with their default value makes that we lose these default value in the table of the properties of the story :(

stories/atoms/components/textField/TextField.stories.mdx
```diff
@@ -266,13 +266,22 @@ The `withBorder` property allows you to add a themed border to the textField and
 
 ## Multiline
 
-The `multiline` property allows you to write text on several lines. The simple text field becomes a text area.  
-When `multiline` is `true` and [disableAreaResize](/docs/atoms-textfield--disable-area-resize) is `false` the [size](/docs/atoms-textfield--sizes) is automatically `auto`.
+The `multiline` property allows you to write text on several lines. The simple text field becomes a text area.
+The [`size`](/docs/atoms-textfield--sizes) is automatically **auto**.  
+When `fullWidth` is **true** the user can only resize the height of the textField.
```
lolottetheclash, line 271: Maybe you could rework this description to have fewer very short sentences and a more airy text with possibly some line breaks?

stories/atoms/components/textField/TextField.stories.mdx
```diff
@@ -266,13 +266,22 @@ The `withBorder` property allows you to add a themed border to the textField and
...
+    <div
+      style={{
+        display: 'flex',
+        flexDirection: 'column',
+        gap: '10px',
+        alignItems: 'center'
+      }}
+    >
+      <TextField placeholder="I'm a multiline text field." multiline />
+      <TextField placeholder="I'm a multiline text field in full width." multiline fullWidth />
```
lolottetheclash, line 284: Also, you could add some indications in the TextField placeholder.
You could take advantage of this to write on several lines, and indicate for example:
"I'm a multiline text field: My size is automatically set to auto & you can resize my width and height!
This is an idea, I let you see what you want to do for both TextField examples.

lolottetheclash (approved):
This is easier to understand, nice refactor ! 👍

fredericvilcot (approved):
Great, thank you 😌

src/ui/atoms/textField/TextField.tsx
```diff
@@ -16,52 +16,28 @@ export type TextFieldProps = InputBaseProps & {
...
 export const TextField: React.FC<TextFieldProps> = ({
   size = 'medium',
   helperText,
-  fullWidth = false,
-  multiline = false,
-  numberOfLines,
-  disableAreaResize = false,
   ...props
 }: TextFieldProps): JSX.Element => {
+  const { multiline, disableAreaResize, fullWidth, hasError }: InputBaseProps = props
```
ErikssonJoakim, line 26: Why not pass these props as arguments?

ErikssonJoakim (approved):
Looking good to me 😄 I've just left you a question I'm curious to know

src/ui/atoms/textField/TextField.tsx
JoeyDjr: Because we don't want to repeat the props of `InputBase`

PR #499, opened 2022-09-02
Refactor/task domain events & commands

fredericvilcot:
This PR adds a massive refactor to the `task` domain, follows https://github.com/okp4/ui/pull/476.

The idea is to implement some principles inherited both from Event Sourcing and CQRS patterns.

In particular, it highlights the necessity to enable events snapshots to mutate the state and to trigger those events by commands that can be invoked by any controller.

When the command is invoked, this is its responsibility to ensure that payloads are well designed and uncorrupted, and this in order to preserve the unicity, immutability and infallibility of the generated event that will create or update the state.

Consequently, it allows to protect entities in the domain and prevent from leaking them through commands, events or selectors.

The next step will be to add a command bus that will execute asynchronous operations and will allow to decorrelate commands from redux.

PR #535, opened 2022-09-15
Refactor/file domain events + helpers

fredericvilcot:
👉 This PR brings some little refactoring to the `File` domain, following the work achieved for the `Task`. (https://github.com/okp4/ui/pull/499)

🧠 It consists mainly in:

- removing entities handling in redux through actions and reducer.
- creating specific commands and update `StoreFiles`
- creating specific events (Event Sourcing) witch embeds now the `actionCreators` files that are moved from the `usecase` folder
- updating use-cases by improving the way errors are handled to preserve events unicity and incorruptibility
- adding syntactic sugar to enhance test assertions for `eventBus` `publish` parameters by moving methods from `task` domain to a `common` folder because they are shared between all the SUT.
- updating the command parameter for the `FilePicker` component when dispatching a `removeFile` command
- updating all test suites impacted by those changes

✅ No induced breaking changes


ccamel (approved):
LGTM! One step further.

lolottetheclash (approved):
Great Job !!! 🚀
I'm not always sure to understand why you keep variable names with or without `payload` ;)
e.g: `StoreFilePayload` to `StoreFile` which is used like this `(storeFilesPayload: DeepReadonly<StoreFiles>)` and some stays with it like `FileRemovedPayload`.

PR #536, opened 2022-09-16
fix: change immutable list name import

lolottetheclash:
When importing the `List` component into an application, we observed that the export of the `List` component collided with the export of the `List` of `Immutable`.

Here is the type before the fix:  

Capture d’écran 2022-09-16 à 15 51 03

Indeed, rollup exported the `List` of the `Immutable` library to the detriment of the `List` component.

So we renamed each import of the `List` from `Immutable` as `ImmutableList` in order to find the export of the `List` component in the build and to be able to use it in other applications.

Here is the correct type of the `List` component after the fix:

Capture d’écran 2022-09-16 à 15 50 31

### DIFF
--- a/src/adapters/faucet/primary/eventListeners.spec.ts
+++ b/src/adapters/faucet/primary/eventListeners.spec.ts
@@ -1,4 +1,4 @@
-import { List } from 'immutable'
+import { List as ImmutableList } from 'immutable'
 import { EventBus } from 'ts-bus'
 import type { BusEvent } from 'ts-bus/types'
 import type { ReduxStore } from 'domain/faucet/store/store'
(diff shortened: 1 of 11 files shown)


fredericvilcot (approved):
Great, good catch! 👀

JoeyDjr (approved):

fredericvilcot:
@ccamel 
Is this OK for you?

ccamel:
> @ccamel
> Is this OK for you?

Not quite. But I presume I lack context.

ErikssonJoakim (approved):
Good catch 👍

lolottetheclash:
> > @ccamel
> > Is this OK for you?
> 
> Not quite. But I presume I lack context.

@ccamel I haven't found any workaround yet and we need to use the list to start the "Explore" page of the portal.
I suggest that we use it like that until we find a more solid fix...

ccamel (approved):

ccamel:
> @ccamel I haven't found any workaround yet and we need to use the list to start the "Explore" page of the portal. I suggest that we use it like that until we find a more solid fix...

Ok. Let's go!

PR #542, opened 2022-09-19
ci: checkout out repository before checking new dependencies

nperon:
This PR fixes the report new dependencies step in the ci workflows.

As a result an output like the following is to be expected in the PR page:

hiwelo_action
This is when a new item called 'smallest' version 1.0.1 is added to the package.json dependencies section.

**Note**
We additionaly experimented another action labelled lirantal/github-action-new-dependencies-advisor@v1.1.1 
The latter adds metrics on the reliability and quality of each new dependency:
lirantal_action
The [liranthal github action](https://github.com/lirantal/github-action-new-dependencies-advisor) appears to be under the MIT license although it advertises a commercial logo. 

I am interested in feedback on whether we should choose liranthal or stick to hiwelo/new-dependencies-action.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -86,7 +86,7 @@ jobs:
       - name: Check for new dependencies
         uses: hiwelo/new-dependencies-action@1.0.1
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.OKP4_TOKEN }}
 
   lint-branch-name:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


ccamel (approved):
Okay! Great. +1:

I don't have a strong opinion on whether to adopt another GitHub action or not. The choice has to be made with a bit of hindsight.

fredericvilcot (approved):

amimart (changes requested):
Good catch, you didn't touch it but my only remark is about the token used to post the comment, could you please use the `secrets.OKP4_TOKEN` instead?

nperon:
The right token is used now. Thanks @amimart.

nperon:
I will keep on with the hiwelo action. Snyk metrics can always be requested for free at [this link](https://snyk.io/advisor/) for whatever new npm dependency.

Issue #562, opened 2022-10-05
🧠 Header: v2.0


lolottetheclash:
### **Purpose**

**1 - This version will add the arrows indicating on which navigation item the user is positioned**
**2 - This version will also add navigation sub-menu**

### **Visual**

![Capture d’écran 2022-09-28 à 14 00 13](https://user-images.githubusercontent.com/35332974/192773547-0392c4c9-ea01-4c1c-8697-a8261d8a8b71.png)

![Capture d’écran 2022-09-28 à 14 00 22](https://user-images.githubusercontent.com/35332974/192773550-8564adff-26cf-41f3-b84d-84bfd66b7810.png)

### **Use cases**

A user will now be able to provide a set of sub-menus to facilitate the user experience.

#### **Typology**

It stays as it is: a molecule.

#### **Properties**
The navigationMenu already exists and looks like: 
```
navigationMenu?: JSX.Element []
```

Now it will become:

```
navigationMenu?: NavigationItem []
```

with:
```
type NavigationItem = {
	item: JSX.Element // Main Item visible on the Header
	subItems?: JSX.Element [] // The main item sub menu
        isSelectedFromStart?: boolean // Allows user to define if an item must be selected from start
}
```

#### **State**

N/A

### **Notes**

N/A

PR #563, opened 2022-10-05
Feat: update header to v2

lolottetheclash:
Thank you for reading this [issue](https://github.com/okp4/ui/issues/562) in order to have all the details of this PR.
Some screenshots:

Capture d’écran 2022-10-10 à 11 59 59

Capture d’écran 2022-10-10 à 11 56 31

Capture d’écran 2022-10-10 à 12 00 11

Capture d’écran 2022-10-10 à 12 00 22

### DIFF
--- a/src/ui/molecules/header/Header.tsx
+++ b/src/ui/molecules/header/Header.tsx
@@ -10,7 +10,14 @@ import { Typography } from 'ui/atoms/typography/Typography'
 import { Icon } from 'ui/atoms/icon/Icon'
 import './header.scss'
 
-type NavigationItemsWithIds = DeepReadonly<Map<string, NavigationItem>>
+type NavigationItemsWithId = DeepReadonly<Map<string, NavigationItem>>
+
+export type NavigationMenuProps = {
+  navigation: NavigationItemsWithId
+  onNavItemSelection: (id: string, isSubItem?: boolean) => void
+  selectedNavItem?: string
+  withBurgerMenu?: boolean
+}
 
 export type NavigationItem = {
   menuItem: JSX.Element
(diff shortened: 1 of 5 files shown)


src/ui/molecules/header/Header.tsx
```diff
@@ -32,18 +51,110 @@ const BurgerMenu = ({
...
+}: DeepReadonly<{
+  onSubItemSelection: () => void
+  subMenu: JSX.Element[]
+}>): JSX.Element => {
+  return (
+    <div className="okp4-header-navigation-submenu">
+      {subMenu.map((navItem: DeepReadonly<JSX.Element>, index: number) => (
+        <div
+          className="okp4-header-navigation-submenu-item"
+          key={index}
```
RaphaelEscrig, line 82: Here, the use of the index as a key can cause issues [more info here](https://reactjs.org/docs/lists-and-keys.html).

RaphaelEscrig (changes requested):
Really nice way to add easily the submenus

src/ui/molecules/header/Header.tsx
lolottetheclash: Yes I know, but I really don't see how to handle keys based on jsx elements that don't have ids or other values that I can use....

RaphaelEscrig: Maybe you can use the `short.generate()` function as it's already imported in the file

lolottetheclash: i thought about it, but I think it's useless... check the answer of this : https://stackoverflow.com/questions/52196127/different-ways-to-add-a-key-to-jsx-element-in-loop-in-react

ErikssonJoakim (approved):
Nice work looks good to me 👍

RaphaelEscrig (approved):

JoeyDjr (approved):
LGTM

src/ui/molecules/header/Header.tsx
```diff
@@ -1,11 +1,30 @@
...
+import type { DeepReadonly, UseState } from 'superTypes'
 import { useMediaType } from 'hook/useMediaType'
+import { useOnKeyboard } from 'hook/useOnKeyboard'
+import { useOnClickOutside } from 'hook/useOnClickOutside'
 import { ThemeSwitcher as Switcher } from 'ui/atoms/theme/ThemeSwitcher'
+import { Typography } from 'ui/atoms/typography/Typography'
 import { Icon } from 'ui/atoms/icon/Icon'
 import './header.scss'
 
+type NavigationItemsWithId = DeepReadonly<Map<string, NavigationItem>>
```
fredericvilcot, line 13: As the `Map` keys are ids, I will have named it simply `NavigationItems`

src/ui/molecules/header/Header.tsx
```diff
@@ -1,11 +1,30 @@
...
+
+export type NavigationMenuProps = {
+  navigation: NavigationItemsWithId
+  onNavItemSelection: (id: string, isSubItem?: boolean) => void
```
fredericvilcot, line 17: It will be better to create an alias for id
```ts
type Id = string
```

fredericvilcot (commented):
Great, just minor remarks ...

src/ui/molecules/header/Header.tsx
lolottetheclash: Good idea!

fredericvilcot (approved):
LGTM!

ccamel (approved):

PR #573, opened 2022-10-12
Feat: change mandatory firstElement property to optional

lolottetheclash:
This PR makes the `firstElement` property optional because there was no reason to force the user to provide this property.
The `Header` no longer has any mandatory properties.
🛠 The css has been completely reworked to facilitate the possible extension of the `Header` properties: 
👉 if we plan to add a `Login` element in JSX, this addition will not force us to modify the existing but to add a behavior.


lolottetheclash:
I am closing this PR because it is no longer relevant.

PR #575, opened 2022-10-13
fix: change sprite generator targets

ErikssonJoakim:
This PR fixes the two svg-sprite scripts that were inverted.


lolottetheclash (commented):
Thanks !!!

lolottetheclash (approved):

fredericvilcot (approved):
Good catch @ErikssonJoakim!
It will be great to remove those CLI options in a next future and move them to a dedicated config file in `yaml` format 😌

JoeyDjr (approved):
Good catch 👁

PR #580, opened 2022-10-14
Refactor: update imports usage with aliases

lolottetheclash:
This PR aims to implement the alias imports usage  in [this discussion](https://github.com/okp4/ui/discussions/417).

PR #589, opened 2022-10-22
fix: fix i18n import

ccamel:
Fix (restore) incorrect aliased import introduced by [#580](https://github.com/okp4/ui/pull/580/files#diff-3c486c63778bc259b34e6cb6c8b44f75eab15770a65456b1933f21920733cc27) which no longer imports translations in the domain `error`.

This leads the [faucet-web](https://faucet.okp4.network) (in prod) to display this (for almost all errors): 

image

And the [design system storybook](https://ui.okp4.space) as well.

image

### DIFF
--- a/src/domain/error/index.ts
+++ b/src/domain/error/index.ts
@@ -1,4 +1,4 @@
-import 'i18n/index'
+import './i18n/index'
 export * from './store/builder/store.builder'
 export * from './store/selector/error.selector'
 export * from './usecase/index'


ccamel:
@lolottetheclash @fredericvilcot to be tested carrefully.

fredericvilcot (approved):
Thanks @ccamel 🙏
We need to be more careful about regressions because it can cause some big confusions (e.g Faucet)

lolottetheclash (approved):
well done, sorry for the regression :(

PR #595, opened 2022-10-24
Fix/faucet connection error

fredericvilcot:
This PR aims to fix the way connection errors are caught by the faucet client, by:

- adding a new `FaucetConnectionError` dedicated to the GQLWS failure
- translating error into a readable message (generic)


amimart (approved):

ccamel (approved):
LGTM! Go. :+1:

PR #680, opened 2023-06-05
Feat/faucet recaptcha

ErikssonJoakim:
This PR introduces a captcha component and adds it to the faucet. As a first implementation it only has the google recaptcha v2 as possible captcha.

The component comes with a hook that injects the Google captcha api and removes it once the captcha de-mounts.

Seeing as we can have two recaptchas at the same time we cannot render the widget implicitly as it targets the first element with the `g-recaptcha` class. 

The official doc can be found here : https://developers.google.com/recaptcha/docs/display.


jeremylgn (approved):
:+1:

lolottetheclash (approved):
👌

